PR #12, opened 2022-08-30
Add .gitignore

Sajjon:
Or should `Package.resolved` maybe be under version control...?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.swiftpm
+Package.resolved
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+.DS_Store
 .swiftpm
-Package.resolved
+/.build
+/Package.resolved


amartini51:
@franklinsch or @bitjammer Can you comment on best practices here?

.gitignore
```diff
@@ -0,0 +1,2 @@
+.swiftpm
+Package.resolved
```
franklinsch, line 2: Since the only dependency here is the package plugin, it makes sense to me to ignore the Package.resolved file, to ensure that we're always using the latest version of the plugin.

You might also want to consider ignoring the `/.build` folder, see https://github.com/apple/swift-markdown/blob/main/.gitignore for inspiration.

franklinsch (approved):

amartini51:
@Sajjon I made some suggested changes to the `.gitignore` file here: https://github.com/amartini51/swift-book/commit/475c2fa9

To add that commit to this branch, you can run the following:

```
# Fetch from my fork
% git remote add amartini51 git@github.com:amartini51/swift-book.git
% git fetch --all

# Add my changes
% git checkout gitignore
% git merge --ff-only amartini51/gitignore
% git push
```

.gitignore
amartini51: Changed in a4c4c43c.

Issue #26, opened 2022-09-06
Maybe downgrade swift version to 5.6


ruurd:
Is it possible to downgrade to 5.6 to make it usable for mere mortals that run latest Monterey?

amartini51:
The version of TSPL on Swift.org always shows the latest version of Swift that's available in Xcode or in an Xcode beta.  However, you can see both the Swift 5.6 and Swift 5.7 (beta) versions of TSPL on Apple Books:

<https://books.apple.com/us/book/the-swift-programming-language-swift-5-6/id881256329>

Under an older publication pipeline, we published two versions of TSPL on the web — beta and non-beta.  Maybe we sould turn this GitHub issue into a request to publish both versions again when we use DocC.

rtharston:
I agree with amartini51, it would be nice to have both beta and release docs available.

@ruurd, I understand you are referring to not running a beta (Ventura or Xcode), but in case you weren't aware, you can run Xcode 14 on Monterey as well (and it is now out of beta, if that was holding you back). :)

amartini51:
Discussed again on the forum: https://forums.swift.org/t/72375

To have it written down, we've had the following things available over time:

- When TSPL was hosted on developer.apple.com, there were separate web versions of the book for beta and prerelease, separate ePUB versions in the Apple Books store for beta and prerelease, and an ePUB version on Swift.org for only the latest.
- When TSPL was hosted on Swift.org using its pre-DocC toolchain, there was a single web version for the latest.  There were separate ePUB versions in the Apple Books store for beta and prerelease, and an ePUB version on Swift.org for only the latest. 
- Currently, with TSPL hosted on Swift.org and built using DocC, there is a single web version for the latest.  We're no longer able to build an ePUB version (see #2).

PR #27, opened 2022-09-12
Changed concurrency version back to 5.5 from 5.7.

### LINKED ISSUE #21: Version Compatibility page Swift book has incorrect information about version required for concurrency
## Describe the bug

The second to last paragraph of the Version Compatibility page of the Swift book incorrectly states concurrency requires '5.7 or later'. It should state '5.5 or later'. (My guess is someone did a 'replace all'.)

The same paragraph also says to, "set a deployment target of at least iOS 15, macOS 12, tvOS 15, or watchOS 8.0" which was correct at the time Swift 5.5 was released, but since Xcode 13.2 it is possible to use Swift Concurrency (but not all the new APIs) back to ["macOS Catalina 10.15, iOS 13, tvOS 13, and watchOS 6"](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes).

## To Reproduce

1. Go to [Version Compatibility](https://docs.swift.org/swift-book/GuidedTour/Compatibility.html)
2. Scroll down to second to last paragraph
3. See error

## Expected behavior

State facts.

## Actual behavior

State mix of old facts and new find-and-replace error.

## Screenshots

image

## Environment details

N/A

### Swift.org version/commit hash

Swift 5.7 book (that is all I could find)

## Additional context 

N/A

rtharston:
Updated concurrency deployment target to reflect Xcode 13.2's support for concurrency back to the two earlier versions of the OSes.

Fixes #21.

To keep changes minimal, and keep the paragraph clean and concise, this purposefully doesn't add wording about Xcode 13 and 13.1 only supporting macOS 12 and up, and Xcode 13.2 adding the backwards compatibility, nor any of the history of how that support was added. If someone wants to add wording to that effect that should probably be discussed separately and elsewhere.

### DIFF
--- a/Sources/TSPL/TSPL.docc/GuidedTour/Compatibility.md
+++ b/Sources/TSPL/TSPL.docc/GuidedTour/Compatibility.md
@@ -35,11 +35,11 @@ the following changes are available only to code that uses Swift 5.7 or later:
   For example, `UInt64(0xffff_ffff_ffff_ffff)` evaluates to the correct value
   rather than overflowing.
 
-Concurrency requires Swift 5.7 or later,
+Concurrency requires Swift 5.5 or later,
 and a version of the Swift standard library
 that provides the corresponding concurrency types.
 On Apple platforms, set a deployment target
-of at least iOS 15, macOS 12, tvOS 15, or watchOS 8.0.
+of at least iOS 13, macOS 10.15, tvOS 13, or watchOS 6.0.
 
 A target written in Swift 5.7 can depend on
 a target that's written in Swift 4.2 or Swift 4,
--- a/Sources/TSPL/TSPL.docc/GuidedTour/Compatibility.md
+++ b/Sources/TSPL/TSPL.docc/GuidedTour/Compatibility.md
@@ -35,7 +35,7 @@ the following changes are available only to code that uses Swift 5.7 or later:
   For example, `UInt64(0xffff_ffff_ffff_ffff)` evaluates to the correct value
   rather than overflowing.
 
-Concurrency requires Swift 5.5 or later,
+Concurrency requires Swift 5.7 or later,
 and a version of the Swift standard library
 that provides the corresponding concurrency types.
 On Apple platforms, set a deployment target


rtharston:
Contributing.md says to add @amartini51 and @krilnon, but I don't have the ability to add reviewers to the PR.

rtharston:
@swift-ci please test

amartini51:
@rtharston Thanks for the PR — I've added myself and Kyle as reviewers. Do you know who we should add as a technical reviewer? Or were you involved in building this feature, so you're essentially already the technical reviewer?

rtharston:
@amartini51, this is my first PR here so I'm not sure who would be best to review.
I was not involved in the concurrency feature in Swift, I just updated the book to reflect the changes from the [Xcode 13.2 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes).

"You can now use Swift Concurrency in applications that deploy to macOS Catalina 10.15, iOS 13, tvOS 13, and watchOS 6 or newer."

amartini51:
Looked at this with @krilnon, and the change to OS versions here matches release notes, so I don't think it needs further review.  However, maybe there's a little confusion around how this chapter uses language versions numbers, which we could improve:

> This book describes Swift 5.7, the default version of Swift that's included in Xcode 14. You can use Xcode 14 to build targets that are written in either Swift 5.7, Swift 4.2, or Swift 4.

TSPL describes only the current version of Swift, which has three language modes: Swift 5.7 (sometimes also called "Swift 5" mode), Swift 4.2, and Swift 4.  There isn't a language version for Swift 5.5.

As a quick fix, how about we put the version number back (line 38) and merge this, and then revisit approaches to clarify that we're talking about the language versions supported by the Swift 5.7 compiler?

rtharston:
Hmm... that makes sense. I didn't think about the Swift version in context of the compiler mode.

Okay, I was ready to agree with that plan, but then I reread the section and saw the bullet points right about the paragraph in question. The first bullet mentions Swift 5.1.

Could the concurrency paragraph be moved into those bullets, similar to the 5.1 bullet?

Since we are assuming the reader of the book is using the latest version of Xcode yes it is true the Swift 5 mode, which is Swift 5.7 right now, gets them concurrency but I don't want them to think the current Xcode is the only version that supports it. Fact is concurrency was added with 5.5.

I agree with your point about the compiler mode, I just feel the wording is off, and there is precedent for calling out an earlier version in that first bullet point.

If you'd like to go with the plan to keep 5.7 for now and then discuss the bullet idea elsewhere I'd be fine with that. The wrong deployment target is the greater sin here that needs corrected. We don't want people thinking they can't use concurrency at all because they support older targets!

amartini51:
> The first bullet mentions Swift 5.1.

This chapter definitely needs some rewording for clarity to emphasize that it's about the 5.7 _language mode_ — that bullet is about the Swift runtime.  So that approach won't work when we're talking about Swift 5.5 which is a language _version_.

The need to improve the documentation experience for readers who are on an older version of Xcode is currently tracked by https://github.com/apple/swift-book/issues/26

amartini51 (approved):

amartini51:
@swift-ci Please test.

rtharston:
> > The first bullet mentions Swift 5.1.
> 
> This chapter definitely needs some rewording for clarity to emphasize that it's about the 5.7 _language mode_ — that bullet is about the Swift runtime. So that approach won't work when we're talking about Swift 5.5 which is a language _version_.

I _think_ it is correct to say that async await requires the Swift 5.5 runtime, but I'm not an expert here.

> The need to improve the documentation experience for readers who are on an older version of Xcode is currently tracked by #26

I'll keep an eye on that issue, thanks!

PR #30, opened 2022-09-16
Include example of defer lexical scoping

etcwilde:
This patch adds an example of how the defer statement is lexically scoped and will execute at the end of the scope, not the end of a function.

### DIFF
--- a/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
+++ b/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
@@ -820,7 +820,7 @@ to perform manual resource management such as closing file descriptors,
 and to perform actions that need to happen even if an error is thrown.
 
 The *statements* in the `defer` statement
-are executed at the end of the scope enclosing the `defer`.
+are executed at the end of the scope that encloses the `defer` statement.
 
 ```swift
 func f(x: Int) {
@@ -861,7 +861,8 @@ f(x: 5)
   ```
 }
 
-The `defer` in the `if` statement
+In the code above,
+the `defer` in the `if` statement
 executes before the `defer` declared in the function `f`
 because the scope of the `if` statement ends
 before the scope of the function.
--- a/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
+++ b/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
@@ -819,6 +819,52 @@ This means that a `defer` statement can be used, for example,
 to perform manual resource management such as closing file descriptors,
 and to perform actions that need to happen even if an error is thrown.
 
+The *statements* in the `defer` statement are executed at the end of the scope
+enclosing the `defer`.
+
+```swift
+func f(x: Int) {
+  defer { print("First defer") }
+
+  if x < 10 {
+    defer { print("Second defer") }
+    print("End of if")
+  }
+
+  print("End of function")
+}
+f(x: 5)
+// Prints "End of if"
+// Prints "Second defer"
+// Prints "End of function"
+// Prints "First defer"
+```
+
+
+@Comment {
+  ```swifttest
+  -> func f(x: Int) {
+    defer { print("First defer") }
+
+    if x < 10 {
+      defer { print("Second defer") }
+      print("End of if")
+    }
+
+    print("End of function")
+  }
+  f(x: 5)
+  <- End of if
+  <- Second defer
+  <- End of function
+  <- First defer
+  ```
+}
+
+The `defer` in the `if` statement executes before the `defer` declared in the
+function `f` because the scope of the `if` statement ends before the scope of
+the function.
+
 If multiple `defer` statements appear in the same scope,
 the order they appear is the reverse of the order they're executed.
 Executing the last `defer` statement in a given scope first
--- a/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
+++ b/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
@@ -819,8 +819,8 @@ This means that a `defer` statement can be used, for example,
 to perform manual resource management such as closing file descriptors,
 and to perform actions that need to happen even if an error is thrown.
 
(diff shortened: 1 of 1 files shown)


Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
```diff
@@ -819,6 +819,27 @@ This means that a `defer` statement can be used, for example,
 to perform manual resource management such as closing file descriptors,
 and to perform actions that need to happen even if an error is thrown.
 
+The `defer` statement is lexically scoped, and will execute at the end of the
+scope where it is declared.
+
+```
+func f(x: Int) {
+  defer { print("First Defer") }
```
etcwilde, line 827: lower-case `Defer` here and fix output below.

Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
```diff
@@ -819,6 +819,27 @@ This means that a `defer` statement can be used, for example,
...
+
+  if x < 10 {
+    defer { print("Second defer") }
+    print("End of if")
+  }
+
+  print("End of Function")
```
etcwilde, line 834: lowercase `Function` and fix output below.

Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
```diff
@@ -819,6 +819,31 @@ This means that a `defer` statement can be used, for example,
 to perform manual resource management such as closing file descriptors,
 and to perform actions that need to happen even if an error is thrown.
 
+The `defer` statement is lexically scoped and executes at the end of the scope
+in which it is declared.
```
etcwilde, line 823: Not sure about "declared" here. Statements aren't declared. "scope in which it is stated"?

krilnon: I think you're mostly defining what it means to be lexically scoped here, so rather than "and", you want something like:

> The `defer` statement is lexically scoped; that is, its contents run at the end of the scope that encloses the `defer` statement.

etcwilde: Okay, re-addressed;

> The *statements* in the `defer` statement are executed at the end of the scope
 enclosing the `defer`.

amartini51:
Thanks for adding this.  Some general thoughts:

I agree with Kyle that we're better not introducing the term "lexical scope" here.  Do you think that's something a developer is likely to encounter on blog posts on on Swift Evolution discussions?  If so, what we've done elsewhere is something like this:

> The `defer` statement executes at the end of the scope that it appears in.
> This behavior is sometimes called _lexical scoping_.

The three part intro-code-explanation structure is very common in the guide, but not so much in the reference.  What do you think about adding a some detail to the guide?  I could try expanding what you wrote and adding to the the [Error Handling](https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html#ID514) chapter's discussion of `defer`.  I'd be interested if you think there's a good way to add `defer` as early as Control Flow, since it's a general flow-control construct, not tied to error handling.  (For my future reference, the "not just error handling" enhancement is <rdar://34173603>)

As a process suggestion, could you please avoid rebasing and force-pushing to an open pull request?  It makes things more difficult to review.  For example, I started looking at this on Friday, and now it's hard for me to see what changed between that version and the current version.

etcwilde:
> I agree with Kyle that we're better not introducing the term "lexical scope" here. Do you think that's something a developer is likely to encounter on blog posts on on Swift Evolution discussions? If so, what we've done elsewhere is something like this:

I can drop the `lexical scope`. It only seems to show up in SE-0025 and SE-0311, so it isn't terribly common in the evolution proposals. It is used a bit more on the forums, but I don't think it's required for someone just starting out with Swift.

> The three part intro-code-explanation structure is very common in the guide, but not so much in the reference. What do you think about adding a some detail to the guide?

Yeah, it probably makes sense to add a description in the Control Flow part of the guide. It looks like the description of `defer` in Error Handling is pretty minimal and could probably even stay there, just a revisitation to show how it can tie in with error handling, kind of like how `guard` makes a small reappearance.

> As a process suggestion, could you please avoid rebasing and force-pushing to an open pull request? It makes things more difficult to review. For example, I started looking at this on Friday, and now it's hard for me to see what changed between that version and the current version.

Sorry about that, everyone has different processes. Here's the diff between the original PR and the rebased PR: https://github.com/apple/swift-book/compare/2856a799f3d2fd0a679a3bc17b6c02852e47799a..fb2bbba0dd9060c2a0fdc5a58e233f535ec7f2e9

amartini51:
This looks pretty good to me.  I made a commit c2d76bd1 to wrap the text as described in the [TSPL style guide](https://github.com/apple/swift-book/blob/main/Style.md#semantic-line-breaks).

Thanks for adding the swifttest block — that will make backporting this change into the published version faster.

I'd to add some discussion under Control Flow here too, unless you'd prefer to keep this PR focused on just the small addition.  I can push to your branch later this afternoon.

etcwilde:
> I'd to add some discussion under Control Flow here too, unless you'd prefer to keep this PR focused on just the small addition. I can push to your branch later this afternoon.

Whatever works. I'm happy as long as there is something describing when defers fire that is a little more interesting than just the function scope.

amartini51:
I've sketched some additions to the guide, but there's enough work to do there that I don't want to hold this PR waiting for it.  I'll push those changes on a separate branch.

amartini51:
@swift-ci Please test.

amartini51 (approved):

Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
etcwilde: @krilnon, are you okay with the changes?

krilnon: Yes, looks good now.

krilnon (approved):

PR #32, opened 2022-09-17
Remove claim on possible optional bindings in repeat-while conditions

SimplyDanny:
The claim is immediatly disproved on the grammer rule description below that states that the `while` condition is an expression. An expression in turn does not comprise an optional binding. The grammar rule seems to be correct since it matches with the bahavior of the Swift compiler and the SwiftSyntax library.

### DIFF
--- a/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
+++ b/Sources/TSPL/TSPL.docc/ReferenceManual/Statements.md
@@ -174,8 +174,6 @@ the *statements* in a `repeat`-`while` statement are executed at least once.
 
 The value of the *condition*
 must be of type `Bool` or a type bridged to `Bool`.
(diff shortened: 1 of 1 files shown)


SimplyDanny:
This is all about [repeat-while statements](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#grammar_repeat-while-statement) which only allow expressions in their while-condition as the grammar states and the compiler expects. An example from a Swift playground:
Screenshot 2022-09-19 at 23 22 01

amartini51 (approved):

amartini51:
Thanks for that context — looking at the diff, I found the wrong part of the book.  This looks good to me!

@swift-ci Please test.

krilnon (approved):

PR #35, opened 2022-09-22
Rename images so they render properly at 2x scale.

krilnon:
The renames update the image names so they follow DocC's rules for 2x images: https://developer.apple.com/documentation/docc/image

Previously files were named like `someImage_2x.png`, now they're `someImage@2x.png` so DocC picks up that it's a 2x image. 

Updates all the image references to remove suffixes.

Before:
Previous large image behavior

After:
Correct smaller image behavior


amartini51 (approved):
Looks good to me, having compared the output.

As currently published:

published

Using DocC, before this change:

docc before

Using DocC, with this change:

docc after

amartini51:
@swift-ci Please build.

PR #42, opened 2022-10-05
Types/type annotation

milenaalcantara:
translated Type Annotation section of Types topic


amartini51:
This PR looks like it should be merging changes into the [Portuguese translation](https://github.com/AcademyIFCE/swift-book) repository.  This is the repository for the original English text.

Issue #4, opened 2022-08-24
Improve the presentation of syntax outlines


krilnon:
When introducing new syntax for the first time, the book uses placeholder literals to show the customizable portion(s) of the syntax being introduced. Right now, these render as regular Swift code blocks that don't have a special treatment for placeholders.

rdar://64038462

PR #44, opened 2022-10-10
Mark "syntax outline" code listings as Swift

Kyle-Ye:
- Update code snippet to Swift
Fix some Swift code snippet does not use `swift` metadata

> Actually there are other cases for this situation. Since they are incomplete Swift Code, should we mark the `swift` metadata to add syntax highlighting support?

Before (Without Swift Syntax highlighting)
image

After
image

- Update gitignore
Ignore the .docc-build folder after `docc preview TSPL.docc`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .swiftpm
 /.build
 /Package.resolved
+.docc-build
--- a/Sources/TSPL/TSPL.docc/LanguageGuide/Closures.md
+++ b/Sources/TSPL/TSPL.docc/LanguageGuide/Closures.md
@@ -134,7 +134,7 @@ using closure expression syntax.
 
 Closure expression syntax has the following general form:
 
-```
+```swift
 { (<#parameters#>) -> <#return type#> in
    <#statements#>
 }
--- a/Sources/TSPL/TSPL.docc/LanguageGuide/CollectionTypes.md
+++ b/Sources/TSPL/TSPL.docc/LanguageGuide/CollectionTypes.md
@@ -192,7 +192,7 @@ which is a shorthand way to write one or more values as an array collection.
 An array literal is written as a list of values, separated by commas,
 surrounded by a pair of square brackets:
 
-```
+```swift
 [<#value 1#>, <#value 2#>, <#value 3#>]
 ```
 
@@ -1142,7 +1142,7 @@ the key and value in each key-value pair are separated by a colon.
 The key-value pairs are written as a list, separated by commas,
 surrounded by a pair of square brackets:
 
-```
+```swift
 [<#key 1#>: <#value 1#>, <#key 2#>: <#value 2#>, <#key 3#>: <#value 3#>]
 ```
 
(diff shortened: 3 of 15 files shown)


krilnon:
Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.

Kyle-Ye:
> Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.

+1. Maybe we need add such feature in Swift-DocC & Swift-DocC-Render.

amartini51:
@swift-ci Please test.

Kyle-Ye:
> > Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.
> 
> +1. Maybe we need add such feature in Swift-DocC & Swift-DocC-Render.

IMO we need to introduce a new structure to capture this in Swift Markdown. What's your opinion on this? cc @QuietMisdreavus 

And then emit it from Swift-DocC to Swift-DocC-Render/Xcode.

The Swift-DocC-Render's implementation detail could follow the current behavior of swift-book website. (Xcode's documentation support should be implemented in Apple and is not the scope of the open source topic cc @franklinsch )

image

QuietMisdreavus:
> > > Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.
> > 
> > 
> > +1. Maybe we need add such feature in Swift-DocC & Swift-DocC-Render.
> 
> IMO we need to introduce a new structure to capture this in Swift Markdown. What's your opinion on this? cc @QuietMisdreavus

Swift-Markdown would be the right place for this, however IIRC it would be a breaking change. As a data structure, code blocks currently store their contents as plain text, not as a collection of inner structures. I do think it would be a good addition, so we should file the issue(s) and get that working.

amartini51:
I think this is an improvement over no syntax highlighting.  However, for the sake of consistency, we should either change all of the syntax outlines in the reference, or leave them all as-is.  The fact that they don't have a language marked was probably an oversight in the RST-to-markdown export process.  @Kyle-Ye ok if I push more commits to this branch?  I don't want to conflict with work you might still be doing.

> Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.

The tools and content changes to render `<# #>` as placeholder tokens can happen on its own PR.

Kyle-Ye:
> I think this is an improvement over no syntax highlighting. However, for the sake of consistency, we should either change all of the syntax outlines in the reference, or leave them all as-is. The fact that they don't have a language marked was probably an oversight in the RST-to-markdown export process. @Kyle-Ye ok if I push more commits to this branch? I don't want to conflict with work you might still be doing.
> 
> > Ideally these would render with a presentation similar to the screenshot in #4, rather than as normal Swift code.
> 
> The tools and content changes to render `<# #>` as placeholder tokens can happen on its own PR.

Yes, you are free to push more commits on this branch.

amartini51:
Thanks, @Kyle-Ye.  I think I've got the rest of them in commit 45717bf1.

amartini51:
@swift-ci Please build.

amartini51 (approved):

krilnon (approved):

PR #49, opened 2022-10-13
Tradução: Statements.md - Control Transfer Statements

daneccc:


krilnon:
I think you want to make this pull request against the [Portuguese translation](https://github.com/AcademyIFCE/swift-book) repository instead. See a similar issue in #42, looks like it was merged over there as this PR: https://github.com/AcademyIFCE/swift-book/pull/574.

Issue #34, opened 2022-09-21
Publish a preview/staging version of the book


krilnon:
Per discussion in the docs workgroup, it'd be good to have a preview version of the DocC output that we can all look at as we work toward publishing the book on swift.org. This preview should be:

- Published on GitHub pages, probably using [a workflow like this one from swift-docc-plugin](https://github.com/apple/swift-docc-plugin/blob/main/bin/update-gh-pages-documentation-site).
- Visually distinguishable as a draft/preview version so we don't confuse people who are looking for the docs and not participating in the DocC conversion effort.

rdar://101885227

krilnon:
To publish this either in a preview environment or on swift.org, we need custom header support. That exists with `--experimental-enable-custom-templates`, but it doesn't work when you also build with `--transform-for-static-hosting`, which we use for TSPL. 

Tracked by https://github.com/apple/swift-docc/issues/196, and I have an initial fix drafted in https://github.com/apple/swift-docc/pull/410.

PR #73, opened 2022-12-01
Add custom header for staging copy and build script

krilnon:
As discussed in some of the docs workgroup meetings, it's useful for us to have a staging copy of the book.

Fixes https://github.com/apple/swift-book/issues/34

### DIFF
--- a/bin/update-book-preview
+++ b/bin/update-book-preview
@@ -1,5 +1,5 @@
-#!/bin/bash
-#
+#! /bin/bash
+
 # This source file is part of the Swift.org open source project
 #
 # Copyright (c) 2022 Apple Inc. and the Swift project authors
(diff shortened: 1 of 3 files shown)


bin/update-book-preview
```diff
@@ -0,0 +1,74 @@
...
+# Check out the 'gh-pages' branch in a subdirectory.
+git worktree add --checkout gh-pages "$REMOTE"/gh-pages
+
+# Pretty print DocC JSON output, so that it has a stable ordering and
+# can be diffed.  This lets us determine whether rebuilding resulted in
+# any changes to the content.
+export DOCC_JSON_PRETTYPRINT="YES"
+
+# Build documentation, writing output in the gh-pages/docs subdirectory.
+export SWIFTPM_ENABLE_COMMAND_PLUGINS=1
```
ethan-kusters, line 46: Nit: This shouldn't be needed anymore.

```suggestion
```

bin/update-book-preview
```diff
@@ -0,0 +1,74 @@
...
+swift package \
+    --allow-writing-to-directory "./gh-pages/docs" \
+    generate-documentation \
+    --target TSPL \
+    --disable-indexing \
+    --transform-for-static-hosting \
```
ethan-kusters, line 52: Nit: Static hosting transformation happens by default now.

```suggestion
```

ethan-kusters (approved):

mportiz08 (approved):
The custom header looks good from a functional perspective 👍

I have some minor suggestions on the styling but I think I should leave that for a separate PR :)

amartini51:
> I have some minor suggestions on the styling but I think I should leave that for a separate PR :)

@mportiz08 If it's minor feedback, probably just as easy for us to discuss and incorporate style changes on this PR.

bin/update-book-preview
amartini51: Removed it 0d56de46.

bin/update-book-preview
amartini51: Removed it 0d56de46.

amartini51:
Now that we're adding a custom footer too (#88) we also need to add some kind of switching mechanism so the header gets used only for previews.  Without the footer, we were ok only setting `--experimental-enable-custom-templates` for previews.

PR #91, opened 2023-01-17
remove comma for correlative conjunction

bomyuniverse:
Hello, I'm Bomi and learning a swift.
While studying, I found a sentence that would be better to modify so I suggest it.

I think the comma should be removed from this sentence so the conjunctions are not separated from each other.
Because `Either...or` is a correlative pair of conjunction.

I couldn't seen using commas in sentences of this chapter include correlative conjunction. I guess it would be better to change it for the uniformity.

Please review, Thanks! 💙


bomyuniverse:
@swift-ci Please test. 😄

bomyuniverse:
Please review 😄

amartini51 (approved):

PR #94, opened 2023-01-24
Add a build script for publication and use a plain DocC bundle

krilnon:
Adds a separate script that generates a version of the book suitable for publishing on swift.org (with the non-staging header).

This conditionalizes which `header.html` is used.

This PR also moves away from using the package plugin to generate the book and instead uses plain DocC commands. This simplifies the file structure of the repo.


krilnon:
We'll need to change the swift-ci config after this change:

```shell
swift package plugin generate-documentation --target TSPL --transform-for-static-hosting
```

…to:

```shell
docc convert --experimental-enable-custom-templates --hosting-base-path swift-book --output-path "./swift-book" TSPL.docc
```

amartini51 (approved):

bjlanier (approved):

PR #110, opened 2023-03-06
Add appearance selection input to the footer

krilnon:
Several people pointed out that they would like a site-specific toggle for light/dark mode for the book: https://forums.swift.org/t/light-mode-for-swift-5-8-docs/63364

This PR adds a toggle like the one in the footer on the main swift.org site. I manually expanded the SCSS to CSS to avoid adding a build-time dependency for it. 

I deployed a test/preview version here: https://krilnon.github.io/swift-book/documentation/the-swift-programming-language/

### DIFF
--- a/TSPL.docc/footer.html
+++ b/TSPL.docc/footer.html
@@ -53,11 +53,11 @@ See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
         outline: none;
         --toggle-color-fill: var(--color-button-background);
         font-size: 12px;
-        border: 1px solid var(--color-nav-link-color);
-        border-radius: var(--border-radius);
+        border: 1px solid #fff;
+        border-radius: 4px;
         display: inline-flex;
         padding: 1px;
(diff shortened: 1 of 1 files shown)


mportiz08:
Haven't had time to look through the code yet, but it seems like the "Auto" selection doesn't change with the system appearance anymore using the preview site that you linked.

krilnon:
Oh, interesting. I think it respects Auto if that's your selection when you load the page, but not if you swap off of Light or Dark and then change your system setting.

amartini51 (approved):
Tested the behavior of the staged version, and it seems to be working properly. Also spot checked that images are respecting light/dark mode.  Also confirmed that the issue from an earlier iteration, where the light/dark/auto switch was bumping into another element in the footer, has been fixed.

Only skimmed the HTML, CSS, and JavaScript here because they were copied from what Swift.org uses.

EDIT: Kyle noted that the HTML etc. is derived from Swift.org's code, but isn't a literal copy of it.

PR #127, opened 2023-04-02
Fix release date error

bbiguduk:
I know that Swift 5.8 release date is Mar 30. But wrong date is displayed to swift-book revision history.
Please check this!
Thank you.


amartini51 (approved):
You're correct - this typo got overlooked when reviewing the 5.8 version.

PR #128, opened 2023-04-03
Add a forward reference about extensions

bytedriver:
This PR inserts a sentence that explains an extra link about `extension` in the section [Preventing Overrides](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/inheritance#Preventing-Overrides) of the chapter "Inheritance".
"Extension" is mentioned here for the first time. Therefore, it is appropriate to attach a link to it.

To be honest, it is not the first time. You might already find some examples in the Swift Tour, and the chapter "The Basics".
Nonetheless, this time it is mentioned without the link in advance of the chapter "Extensions".
As such, it would be better to have a link.

### DIFF
--- a/TSPL.docc/LanguageGuide/Inheritance.md
+++ b/TSPL.docc/LanguageGuide/Inheritance.md
@@ -510,7 +510,7 @@ Any attempt to override a final method, property, or subscript in a subclass
 is reported as a compile-time error.
 Methods, properties, or subscripts that you add to a class in an extension
 can also be marked as final within the extension's definition.
-For more information about extensions, see <doc:Extensions>.
+For more information, see <doc:Extensions>.
 
 <!--
   - test: `finalPreventsOverriding`
--- a/TSPL.docc/LanguageGuide/Inheritance.md
+++ b/TSPL.docc/LanguageGuide/Inheritance.md
@@ -510,6 +510,7 @@ Any attempt to override a final method, property, or subscript in a subclass
 is reported as a compile-time error.
 Methods, properties, or subscripts that you add to a class in an extension
 can also be marked as final within the extension's definition.
+For more information about extensions, see <doc:Extensions>.
 
 <!--
   - test: `finalPreventsOverriding`


amartini51:
Thanks for looking at this, @bytedriver!

Looking through the places called out above:

- In the tour, it's an intentional decision that we don't link forward to each topic that's mentioned, so it's correct not to have a link.
- In The Basics > [Integer Conversion](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/thebasics/#Integer-Conversion) we have a forward reference when extensions are discussed as a way to add make additional types convertible to `Int`
- In Inheritance > [Preventing Overrides](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/inheritance/#Preventing-Overrides), extensions are mentioned by name.  I'm not entirely sure that there needs to be a forward reference here, because the discussion is so limited and there's a chapter named Extensions.  But if we include one, I'd suggest this wording:

> Methods, properties, or subscripts that you add to a class in an extension can also be marked as final within the extension’s definition.  For more information, see Extensions.

bytedriver:
Thank you for reviewing it again @amartini51 
I made a little change to the description as you suggested.
👍

amartini51:
I think this follow the existing precedent of forward references in the book.  Looking at some selected examples:

```
% git grep --context=2 'doc:[a-zA-Z]*>'

BasicOperators.md-This chapter describes the common operators in Swift.
BasicOperators.md:<doc:AdvancedOperators> covers Swift's advanced operators,
BasicOperators.md-and describes how to define your own custom operators
BasicOperators.md-and implement the standard operators for your own custom types.

CollectionTypes.md-> Note: Swift's array, set, and dictionary types are implemented as *generic collections*.
CollectionTypes.md:> For more about generic types and collections, see <doc:Generics>.

CollectionTypes.md:> For information about conforming to protocols, see <doc:Protocols>.

ErrorHandling.md-For more information about pattern matching,
ErrorHandling.md:see <doc:Patterns>.

Initialization.md:For more information about deinitializers, see <doc:Deinitialization>.

Initialization.md-Classes, however, can inherit from other classes,
Initialization.md:as described in <doc:Inheritance>.

Protocols.md:For information on optional chaining, see <doc:OptionalChaining>.
```

So there are plenty of places where we say "for information about _topic_, see _Chapter_" even when the topic and chapter are the same.

PR #141, opened 2023-06-03
Edits to punctuation to conform to style guide

dmvj:
Fixing punctuation around code examples to conform to the style guide. Only punctuation has changed.

### DIFF
--- a/TSPL.docc/LanguageGuide/TheBasics.md
+++ b/TSPL.docc/LanguageGuide/TheBasics.md
@@ -87,7 +87,7 @@ The current login attempt counter is declared as a variable,
 because this value must be incremented after each failed login attempt.
 
 You can declare multiple constants or multiple variables on a single line,
-separated by commas:
+separated by commas.
 
 ```swift
 var x = 0.0, y = 0.0, z = 0.0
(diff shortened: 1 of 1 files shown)


xwu:
I rather like that code examples are introduced by colons; this seems the more intuitive usage for me as a reader. Can we conform the style guide to the actual usage instead?

dmvj:
> Can we conform the style guide to the actual usage instead?

That's exactly why I [posted over at the Swift Forums](https://forums.swift.org/t/copyedits-of-tspl/65341). It doesn't matter to me either way. I did notice, though, that newer additions to sections tended to conform more closely to the present style guide. So, either way, there's some cleanup to be done. I'm happy to go in either direction. I just need to know which one. :)

amartini51:
The following comments are from Susan (one our editors), who reviewed the PR today:

The punctuation/wording, etc is fine, except for these few instances where colons are needed below:

The following example reports a compile-time error.

```swift
let i = 1
```

Should be: 

The following example reports a compile-time error: (use a colon because the sentence both directly references and precedes the code snippet.)

```swift
let i = 1
```



And the others (changed here):

However, the alternative example below is valid:

```swift
let i = 1
```

The example below uses the initializer to try to convert a String into an Int:

```swift
let possibleNumber = "123"
```



Write an optional binding for an if statement as follows:

```swift
if let <#constantName#> = <#someOptional#> {
```



The following if statements are equivalent:

```swift
if let firstNumber = Int("4"), let secondNumber = Int("42"), firstNumber < secondNumber && secondNumber < 100 {
```


The rubric is: ONLY use colons when the sentence preceding the code snippet directly references that code;  otherwise use a period.

PR #147, opened 2023-06-30
Use the short syntax to unwrap optional properties

ofirmiron:
Because we're reusing the property names, we can use the shorter form as described in The Basics.

### DIFF
--- a/TSPL.docc/LanguageGuide/OptionalChaining.md
+++ b/TSPL.docc/LanguageGuide/OptionalChaining.md
@@ -352,7 +352,7 @@ class Address {
         var buildingNumber: String?
         var street: String?
         func buildingIdentifier() -> String? {
-           if let buildingNumber = buildingNumber, let street = street {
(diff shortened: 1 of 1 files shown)


amartini51:
Edited the PR's title to avoid confusion with implicitly unwrapped optionals.

amartini51:
Thanks for the PR!  This change looks good to me.  I confirmed that this is the only code listing in the chapter that could switch over to the newer short syntax.

amartini51 (approved):

ofirmiron:
Thanks, done!

amartini51:
> Thanks, done!

Not sure what happened, but your edited commit didn't show up.  I still saw https://github.com/apple/swift-book/pull/147/commits/8456accb6ce3234da3bf0f12a3108377bd48a52d

Since you're ok with the change, I amended the commit to edit its message, and force pushed it here.

PR #148, opened 2023-07-05
Fix incorrect parameter names for buildPartialBlock(accumulated:next:)

gwynne:
Fixed two locations where `buildPartialBlock(accumulated:next:)` was incorrectly referred to as `buildPartialBlock(second:)`.

_Note: I did make sure that neither was actually an incorrectly-named reference to `buildEither(second:)`._


amartini51 (approved):

amartini51:
Spot checked these changes against SE-0348.  This looks like a copy/paste error that was overlooked in PR #120.

PR #149, opened 2023-07-05
Add missing `open` access control level examples

ofirmiron:
The `open` access control level was missing from the Access Control Syntax section.


amartini51 (approved):

amartini51 (approved):
Looked over the diff and the [current version](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/accesscontrol/#Access-Control-Syntax).  I agree that it makes sense to show `open` in the list of syntax examples.

Issue #140, opened 2023-05-27
The Result Builders section should be moved out of the Advanced Operators chapter and into its own chapter


WowbaggersLiquidLunch:
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/advancedoperators#Result-Builders

### Description

Because result builders are not operators. 

Relevant forums topic before TSPL was open-source: https://forums.swift.org/t/move-the-result-builders-section-in-tspl-to-its-own-chapter/50056

amartini51:
Where do you suggest moving it?

When I added this section, we had a long discussion about where to put this content.  There isn't anyplace in TSPL that's an exact fit.  Part of the thinking at the time was that we could move the Advanced Operators chapter towards more of a general scope to include DSLs.

WowbaggersLiquidLunch:
Instead of fitting it with something else, why not just have Result Builders be its own chapter? Is the discussion about fitting it somewhere somewhat about its place in the progressive disclosure of complexity?

krilnon:
I'm a fan of turning Advanced Operators into something more of an Advanced Syntax (or similar) chapter. 

Seems like the argument against putting result builders in their own chapter is that there perhaps isn't enough to say about them to warrant a whole chapter?

WowbaggersLiquidLunch:
> Seems like the argument against putting result builders in their own chapter is that there perhaps isn't enough to say about them to warrant a whole chapter?

Judging by the length of the proposal document, there shouldn't be a shortage of things to say about it. And even if there isn't much, it can just be a short chapter, like the _Nested Types_ chapter.

WowbaggersLiquidLunch:
Just noticed that some of Result Builders' documentation is incorrectly put [in the _Attributes_ chapter](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes#resultBuilder).

PR #151, opened 2023-07-06
Describe how result builders interact with "build partial block"

gwynne:
The `buildPartialBlock(first:)` and `buildPartialBlock(accumulated:next:)` methods are described by the "Result-Building Methods" subsection of `@resultBuilder`, but there is no corresponding detail provided in the "Result Transformations" subsection. This adds the appropriate description, including example code which demonstrates the intended use case.

(The point is admittedly partly moot in Swift 5.9, given that it became possible to write a result builder that has `func buildBlock<each D: Drawable>(_ components: repeat each D) -> some Drawable`, but the `buildPartialBlock()` methods may still be a better fit for some use cases. Even if this were not the case, they still exist and should be fully documented.)

### DIFF
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -1883,7 +1883,7 @@ into code that calls the static methods of the result builder type:
       var second: Second
       func draw() -> String { return first.draw() + second.draw() }
   }
-  
+
   @resultBuilder
   struct DrawingPartialBlockBuilder {
       static func buildPartialBlock<D: Drawable>(first: D) -> D {
@@ -1895,7 +1895,7 @@ into code that calls the static methods of the result builder type:
           return DrawBoth(first: accumulated, second: next)
       }
   }
-  
+
   @DrawingPartialBlockBuilder var builderBlock: some Drawable {
       Text("First")
       Line(elements: [Text("Second"), Text("Third")])
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -1866,7 +1866,84 @@ into code that calls the static methods of the result builder type:
     >> assert(builderOptional == manualOptional)
     ```
   -->
-- A code block or `do` statement
+- If the result builder implements the `buildPartialBlock(first:)` and
+  `buildPartialBlock(accumulated:next:)` methods, a code block or `do`
+  statement becomes a call to those methods. The first statement inside
+  of the block is transformed to become an argument to the
+  `buildPartialBlock(first:)` method, and the remaining statements
+  become nested calls to the `buildPartialBlock(accumulated:next:)`
+  method. For example, the following declarations are equivalent:
+
+  ```swift
+  struct DrawBoth<First: Drawable, Second: Drawable>: Drawable {
+      var first: First
+      var second: Second
+      func draw() -> String { return first.draw() + second.draw() }
+  }
+  
+  @resultBuilder
+  struct DrawingPartialBlockBuilder {
+      static func buildPartialBlock<D: Drawable>(first: D) -> D {
+          return first
+      }
+      static func buildPartialBlock<Accumulated: Drawable, Next: Drawable>(
+          accumulated: Accumulated, next: Next
+      ) -> DrawBoth<Accumulated, Next> {
+          return DrawBoth(first: accumulated, second: next)
+      }
+  }
+  
+  @DrawingPartialBlockBuilder var builderBlock: some Drawable {
+      Text("First")
+      Line(elements: [Text("Second"), Text("Third")])
+      Text("Last")
+  }
+
+  let partialResult1 = DrawingPartialBlockBuilder.buildPartialBlock(first: Text("first"))
+  let partialResult2 = DrawingPartialBlockBuilder.buildPartialBlock(
+      accumulated: partialResult1,
+      next: Line(elements: [Text("Second"), Text("Third")])
+  )
+  let manualResult = DrawingPartialBlockBuilder.buildPartialBlock(
+      accumulated: partialResult2,
+      next: Text("Last")
+  )
+  ```
+
+  <!--
+    - test: `drawing-partial-block-builder`
+
+    ```swifttest
+    -> @resultBuilder
+    -> struct DrawingPartialBlockBuilder {
+           static func buildPartialBlock<D: Drawable>(first: D) -> D {
+               return first
+           }
+           static func buildPartialBlock<Accumulated: Drawable, Next: Drawable>(
+               accumulated: Accumulated, next: Next
+           ) -> DrawBoth<Accumulated, Next> {
+               return DrawBoth(first: accumulated, second: next)
+           }
+       }
+    -> @DrawingPartialBlockBuilder var builderBlock: some Drawable {
+          Text("First")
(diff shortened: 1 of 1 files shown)


amartini51:
I took a quick formatting pass (892c9931) to make this content use [semantic line breaks](https://github.com/apple/swift-book/blob/main/Style.md#semantic-line-breaks), but I need to look over this more carefully.

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -1866,7 +1866,87 @@ into code that calls the static methods of the result builder type:
...
+  
+  @resultBuilder
+  struct DrawingPartialBlockBuilder {
+      static func buildPartialBlock<D: Drawable>(first: D) -> D {
+          return first
+      }
+      static func buildPartialBlock<Accumulated: Drawable, Next: Drawable>(
+          accumulated: Accumulated, next: Next
+      ) -> DrawBoth<Accumulated, Next> {
+          return DrawBoth(first: accumulated, second: next)
```
krilnon, line 1895:
```suggestion
         DrawBoth(first: accumulated, second: next)
```

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -1866,7 +1866,87 @@ into code that calls the static methods of the result builder type:
...
+  struct DrawBoth<First: Drawable, Second: Drawable>: Drawable {
+      var first: First
+      var second: Second
+      func draw() -> String { return first.draw() + second.draw() }
+  }
+  
+  @resultBuilder
+  struct DrawingPartialBlockBuilder {
+      static func buildPartialBlock<D: Drawable>(first: D) -> D {
+          return first
```
krilnon, line 1890:
```suggestion
         first
```

gwynne:
@krilnon I originally wrote the code exactly according to your suggestions, but I changed it in order to stay consistent with the style of the other code examples in the same section (as I figured changing the style of the existing examples would be arguably out of scope for this PR).

krilnon:
Totally reasonable way to approach it! Thanks for considering.

amartini51:
The prose changes here look good to me.  As a future improvement, along the lines of discussion in https://github.com/apple/swift-book/issues/140 maybe we should consider moving the details of result builders out of the Reference > Attributes chapter — that portion getting rather long, and it has a lot more guide-like example-driven content than the reference typically does.

@gwynne Who should we ask to be a technical reviewer for this change?  I see in the compiler's Git history that you've made many contributions, but it doesn't look like you have worked on the implementation of result builders specifically.  (If that's not the case, please correct me.)

gwynne:
> @gwynne Who should we ask to be a technical reviewer for this change? I see in the compiler's Git history that you've made many contributions, but it doesn't look like you have worked on the implementation of result builders specifically. (If that's not the case, please correct me.) Some possible reviewers include Richard Wei, who wrote the [SE proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0348-buildpartialblock.md), and Pavel Yaskevich, who reviewed the [implementation](https://github.com/apple/swift/pull/41576).

Ah, yeah, that's a good point; I haven't gotten anywhere near result builders at the compiler level - the details I added in this PR came from a combination of language-level experience and reading the compiler source 😅. Pavel would probably be my first choice as far as technical reviewer, with Richard or Doug as backup.

xedin:
Sorry for the delay, I will take a look tomorrow!

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -1866,7 +1866,87 @@ into code that calls the static methods of the result builder type:
...
+- If the result builder implements
+  the `buildPartialBlock(first:)`
+  and `buildPartialBlock(accumulated:next:)` methods,
+  a code block or `do` statement becomes a call to those methods.
+  The first statement inside of the block
+  is transformed to become an argument
+  to the `buildPartialBlock(first:)` method,
+  and the remaining statements become nested calls
+  to the `buildPartialBlock(accumulated:next:)` method.
+  For example, the following declarations are equivalent:
```
xedin, line 1878: Should this section mention what happens when a builder implements both buildPartialBlock and buildBlock but with different availability?

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -1866,7 +1866,87 @@ into code that calls the static methods of the result builder type:
...
+
+  ```swift
+  struct DrawBoth<First: Drawable, Second: Drawable>: Drawable {
+      var first: First
+      var second: Second
+      func draw() -> String { return first.draw() + second.draw() }
```
xedin, line 1884: I'm not sure why previous comment about `return` has been resolved but there is no need for that in all three cases because there is only one expression in the body.

gwynne: > I'm not sure why previous comment about `return` has been resolved but there is no need for that in all three cases because there is only one expression in the body.

[My previous reply](https://github.com/apple/swift-book/pull/151#issuecomment-1630291632) explains why I left this the way it is:

> > @krilnon I originally wrote the code exactly according to your suggestions, but I changed it in order to stay consistent with the style of the other code examples in the same section (as I figured changing the style of the existing examples would be arguably out of scope for this PR).

xedin: I see, since it wasn’t in the thread I assumed it was unrelated…

TSPL.docc/ReferenceManual/Attributes.md
xedin: If partial block is less available than buildBlock the compiler would pick the latter.

gwynne: Is there a real-world use case for these semantics? (Apologies for my ignorance on this one- I spend 99% of my Swift time server-side these days; my only encounters with availability generally have to do with the compiler complaining that a given API requires a newer macOS as I'm trying to target Linux 😅.)

TSPL.docc/ReferenceManual/Attributes.md
gwynne: Sorry for the confusion! IIRC my thought process went "it'd be silly to repeat myself in all three suggestion comments" and somehow it didn't occur to me until later that I could have just replied to any one of them 😅

TSPL.docc/ReferenceManual/Attributes.md
xedin: Yeah, it's intended to cover a use-case where buildPartialBlock is introduced after buildBlock which gives it a way to support both based on the deployment target of the app in question (because i.e. buildBlock is limited in the number of elements it supports).

amartini51: We describe the case where you implement both "build partial block" and "build block" earlier in the chapter:

https://github.com/apple/swift-book/blob/main/TSPL.docc/ReferenceManual/Attributes.md?plain=1#L1501

@xedin Is that what you were thinking of?

xedin: Yeah, that sounds good!

xedin (approved):

TSPL.docc/ReferenceManual/Attributes.md
xedin: No worries!

amartini51 (approved):

bjlanier (approved):

Issue #157, opened 2023-07-18
The defer example in A Swift Tour page produces a warning


ishaanbedi:
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/guidedtour/

### Description

The usage of defer in the example provided on the guided tour page generates a warning. The warning message is as follows:

```
main.swift:15:1: warning: result of call to 'fridgeContains' is unused
fridgeContains("banana")
^             ~~~~~~~~~~
false
```
because the result of the function call to `fridgeContains` is not being used or assigned to a variable.

Although the warning message is somewhat self-explanatory, it could potentially cause confusion among beginners referring to the official guide to Swift.

### Correction

To avoid this, the assignment of the result of the function call to a variable should be done. 

This will ensure that the examples in the official documentation are free of warnings and errors.

I'm ready to do a pull request to solve this after approval.

amartini51:
Please go ahead and open a PR with your suggested change.  This change is small, and doesn't need a pitch on the forums.

For comparison, you can look at commit f4a8a4343abc81e20afc143240fe2b773ec959c7 which also resolves "unused value" warnings.

Before this book moved to DocC, as part of open-sourcing, the tour chapter was available as an Xcode playground.  There  used to be a note box that recommend the playground version as the best experience.  However, DocC isn't able to build the playground version, so it's no longer available.  In a playground, code like this shows `false` as a result without warning about an unused result:

<img width="775" alt="image" src="https://github.com/apple/swift-book/assets/15617351/cca39e3e-495e-417f-9a2b-5b43bb9e6572">

ishaanbedi:
> Is this this the only place you see this issue, or should we check the rest of the tour?

Yes! I followed the guide and this was the only one I found.

ishaanbedi:
Hi @amartini51 !
I've raised a PR for this issue (#158)
Can you please review it?

amartini51:
Tested the existing version in a playground, the Swift REPL, and calling `swift` on the command line with a file.  I see a warning only in the last case.  That's a valid way to read the guided tour, but I wouldn't expect it to be the most common.

Confirmed that the revised version of the code looks ok in all three places.

@ishaanbedi I pushed a commit to your branch with some minor changes.  How does that look to you?

PR #158, opened 2023-07-18
Fix unused result warning for `defer` example

ishaanbedi:
Fix unused result warning for `defer` example in GuidedTour.md for Guided Tour Section

Fixes: https://github.com/apple/swift-book/issues/157

### DIFF
--- a/TSPL.docc/GuidedTour/GuidedTour.md
+++ b/TSPL.docc/GuidedTour/GuidedTour.md
@@ -2227,9 +2227,8 @@ func fridgeContains(_ food: String) -> Bool {
     let result = fridgeContent.contains(food)
     return result
 }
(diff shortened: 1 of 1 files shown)


amartini51:
Tested the existing version in a playground, the Swift REPL, and calling `swift` on the command line with a file.  I see a warning only in the last case.  That's a valid way to read the guided tour, but I wouldn't expect it to be the most common.

Confirmed that the revised version of the code looks ok in all three places.

@ishaanbedi I pushed a commit to your branch with some minor changes.  How does that look to you?

ishaanbedi:
> @ishaanbedi I pushed a commit to your branch with some minor changes. How does that look to you?

Looks good to me. Addresses the issue correctly.

amartini51 (approved):

PR #161, opened 2023-07-23
Unflip two words in Macros chapter

e78l:
And add a period


amartini51 (approved):
These fixes look good — thank you!

PR #162, opened 2023-07-24
Fix misspelling in Opaque Types chapter

natecook1000:
Fixes rdar://112801273 (Type where identify should be identity)


amartini51 (approved):
Thanks!

Issue #163, opened 2023-07-25
Unable to preview site via docc preview TSPL.docc


patternina:
Hi. I'm trying to build the project locally but I'm getting the error below when running `swift run docc preview TSPL.docc` via the Terminal:

```bash
error: Could not find Package.swift in this directory or any of its parent directories.
```

The build instructions do not mention anything related to a `Package.swift` file. Is there anything else I'm supposed to check/install prior to attempting to build the project?

PR #164, opened 2023-07-25
Update README for guiding docc location

Kyle-Ye:
Any wording review comments are welcomed.

https://forums.swift.org/t/unable-to-build-swift-book-via-docc/66342

Fixes: https://github.com/apple/swift-book/issues/163

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ in this repository's root directory.
 >
 > In Swift toolchains downloaded from swift.org, `docc` is located at `/usr/bin/docc` relative to the installation path of the toolchain.
 >
-> If you have Xcode installed on `/Applications`, it will be `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/docc`. You can also use `xcrun docc` to help locate it.
+> If you have Xcode installed, run `xcrun docc preview TSPL.docc`.
 
 After running DocC, open the link that `docc` outputs
 to display a local preview in your browser.
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ in this repository's root directory.
 
 > If you see the error `command not found: docc`, consider adding `docc` to your shell's `$PATH`.
 >
-> Currently, It is locating at a relative path of "/usr/bin/docc" inside your Swift toolchain.
+> In Swift toolchains downloaded from swift.org, `docc` is located at `/usr/bin/docc` relative to the installation path of the toolchain.
 >
 > If you have Xcode installed on `/Applications`, it will be `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/docc`. You can also use `xcrun docc` to help locate it.
 
--- a/README.md
+++ b/README.md
@@ -40,11 +40,17 @@ For more information, see [Contributing to The Swift Programming Language][contr
 Run `docc preview TSPL.docc`
 in this repository's root directory.
 
-> If you see the error `command not found: docc`, consider adding `docc` to your shell's `$PATH`.
->
-> In Swift toolchains downloaded from swift.org, `docc` is located at `/usr/bin/docc` relative to the installation path of the toolchain.
->
-> If you have Xcode installed, run `xcrun docc preview TSPL.docc`.
-
 After running DocC, open the link that `docc` outputs
 to display a local preview in your browser.
+
+> Note:
+>
+> If you installed DocC by downloading a toolchain from Swift.org,
+> `docc` is located in `usr/bin/`,
+> relative to the installation path of the toolchain.
+> Make sure your shell's `PATH` environment variable
+> includes that directory.
+>
+> If you installed DocC by downloading Xcode,
+> run `xcrun docc preview TSPL.docc` instead.
+
--- a/README.md
+++ b/README.md
@@ -40,5 +40,11 @@ For more information, see [Contributing to The Swift Programming Language][contr
 Run `docc preview TSPL.docc`
 in this repository's root directory.
 
+> If you see an error "command not found: docc", you may consider adding `docc` to your $PATH.
+>
+> Currently, It is locating at a relative path of "/usr/bin/docc" inside your Swift toolchain.
+>
+> If you have Xcode installed on `/Applications`, it will be `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/docc`. You can also use `xcrun docc` to help locate it.
+
 After running DocC, open the link that `docc` outputs
 to display a local preview in your browser.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,5 +40,11 @@ For more information, see [Contributing to The Swift Programming Language][contr
 Run `docc preview TSPL.docc`
 in this repository's root directory.
 
+> If you see an error "command not found: docc", you may consider adding `docc` to your $PATH.
```
krilnon, line 43:
```suggestion
> If you see the error `command not found: docc`, consider adding `docc` to your shell's `$PATH`.
```

README.md
```diff
@@ -40,5 +40,11 @@ For more information, see [Contributing to The Swift Programming Language][contr
...
+>
+> Currently, It is locating at a relative path of "/usr/bin/docc" inside your Swift toolchain.
```
krilnon, line 45:
```suggestion
> In Swift toolchains downloaded from swift.org, `docc` is located at `/usr/bin/docc` relative to the installation path of the toolchain.
```

README.md
```diff
@@ -40,5 +40,11 @@ For more information, see [Contributing to The Swift Programming Language][contr
...
+>
+> If you have Xcode installed on `/Applications`, it will be `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/docc`. You can also use `xcrun docc` to help locate it.
```
krilnon, line 47:
```suggestion
> If you have Xcode installed, run `docc preview TSPL.docc`.
```

I think it makes sense to simplify the Xcode instructions since you don't really need to know too many details if you're running via `xcrun`.

krilnon (changes requested):
Seems like a good addition to the instructions in general, since most people won't have `docc` in their path by default. It might be worth telling people about the `xcrun` approach first.

README.md
Kyle-Ye: Is there a typo here? Do you mean
```suggestion
> If you have Xcode installed, run `xcrun docc preview TSPL.docc`.
```

krilnon (approved):
Looks good to me! Seems comprehensive enough that most people should be able to get up-and-running with these additions.

amartini51:
I pushed a bit of a rewrite in a83f76e3 with some more detailed notes about the changes in that commit.

Overall, I'm not entirely sure that it's the README's job to explain how to install DocC.  It might be better if we can find (or write) some documentation in the DocC project about how to install and use `docc` that this README can link to.

README.md
```diff
@@ -42,3 +42,15 @@ in this repository's root directory.
 
 After running DocC, open the link that `docc` outputs
 to display a local preview in your browser.
+
+> Note:
+>
```
amartini51, line 47: If the explicit "Note" feels too heavy, maybe we don't need this in an aside/box at all.  I think it would also work to just include these two sentences at the end of the section.

README.md
```diff
@@ -42,3 +42,15 @@ in this repository's root directory.
...
+> If you installed DocC by downloading a toolchain from Swift.org,
+> `docc` is located in `usr/bin/`,
```
Kyle-Ye, line 49: Nit:

Should this be `/usr/bin` or `usr/bin/`.

IMO, `/usr/bin` is a more common express for relative path.

```suggestion
> `docc` is located in `/usr/bin/`,
```

README.md
Kyle-Ye: For me, this is okay, because I think it is some kind of supplementary information. And it is not particularly important for developers who are familiar with the context. (How to find `docc` executable)

Kyle-Ye (commented):
Thanks for the rewrite.

README.md
amartini51: The leading `/` is for an absolute path.  Relative paths start with `./` or `../` or the name of a file or directory.

I could go either way on the trailing slash.  I included it to emphasize the fact that `bin/` is a directory, but it's also correct to omit it.

README.md
amartini51: I think we should keep it in a note. You mentioned supplementary information, which reminds me that [Apple Style Guide](https://support.apple.com/guide/applestyleguide/n-apsgb2f4a0d6/1.0/web/1.0) calls for Note boxes for information that’s relevant to a topic but may not apply to all readers, or for a bit of information that’s tangential to a topic.

README.md
Kyle-Ye:
```
type id:                    com.apple.dt.document.toolchain (0xeb04)
bundle:                     Xcode (0xed8)
uti:                        com.apple.dt.document.toolchain
localizedDescription:       "en" = ?, "LSDefaultLocalizedValue" = "Xcode Toolchain"
flags:                      active  apple-internal  exported  trusted (0000000000000055)
icons:                      2 values (45420 (0xb16c))
                            {
                                UTTypeIconBadgeName = "hammer-badge";
                                UTTypeIconText = toolchain;
                            }
conforms to:                public.composite-content, com.apple.package
tags:                       .xctoolchain
```

`xctoolchain` conforms to a Uniform Type Identifiers kind of `com.apple.package` which IMO is some kind of small file system. So we should use a absolute path `/usr/bin/` for this small file system. 

> "A single file system object can be a package and a bundle."
> https://developer.apple.com/documentation/uniformtypeidentifiers/uttype/3551543-package

krilnon: I don't have an opinion here either way; if I were speaking aloud to someone trying to convey where the DocC binary lived, I'd say something like "it ships in the Swift toolchain under user bin" and trust that they'd understand the ambiguity.

Kyle-Ye: Actually I'm fine with the current state. (This is just a small suggestion not a "request change")

Is there any other block issues preventing this being merged? @amartini51 @krilnon

amartini51: @Kyle-Ye Thanks for taking the time to make sure this detail is correct. Looking at [Bundle Programming Guide](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/AboutBundles/AboutBundles.html#//apple_ref/doc/uid/10000123i-CH100-SW1), Apple's docs define a package as "any directory that the Finder presents to the user as if it were a single file."  So I prefer to omit the leading `/` and use a relative path.

> A single file system object can be a package and a bundle.

This doesn't mean a package is a kind of file system.  The same directory in the file system can be both a package and a bundle, at the same time.  "Bundle Programming Guide" talks about this in more detail, which maybe should be added to the page you linked:

> The reason bundles and packages are sometimes considered to be interchangeable is that many types of bundles are also packages. For example, applications and loadable bundles are packages because they are usually treated as opaque directories by the system. However, not all bundles are packages and vice versa.

Kyle-Ye: Got it. I'll check the source you provide. And I think we are ready for a merge.

krilnon (approved):

Issue #139, opened 2023-05-26
Show constant with deferred initialization in the guide


amartini51:
### Location

Probably in "The Basics", as part of the discussion of constants.

### Description

Some of the simple cases for this are better expressed as `if` expressions, now that Swift supports them.  However, something like the following is still useful:

```
let x: Int

if someCondition {
    doSomething()
    x = 10
} else {
    doAnotherThing()
    x = 9000
}
```

Using this syntax means you can set the value of `x` depending on a condition.  Writing `x` as a variable above would mean you could accidentally mutate it in later code.

### Motivation

This information is currently only in the reference:

> When a constant declaration occurs in the context of a function or method, it can be initialized later, as long as it’s guaranteed to have a value set before the first time its value is read.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/declarations/#Constant-Declaration

rdar://114058607

PR #172, opened 2023-08-09
Example of constant with deferred assignment

natikgadzhi:
## What

This pull request adds an example of defining a constant without assigning a value to it immediately, and then setting a value based on a condition. 

Fixes: https://github.com/apple/swift-book/issues/139
Fixes: rdar://114058607

## Checklist

- I'm just getting started and learning — I _think_ I got the `swifttest` syntax right? 

/cc @amartini51. How does that look? Are small contributions like this one useful at all? 🙃

### DIFF
--- a/TSPL.docc/LanguageGuide/TheBasics.md
+++ b/TSPL.docc/LanguageGuide/TheBasics.md
@@ -117,8 +117,12 @@ if environment == "development" {
   ```
 -->
 
-In this example, the maximum number of login attemts is set
-to 100 for development environment, or to 10 for any other environments.
+In this example, the maximum number of login attemts is set to 100 for
+development environment, or to 10 for any other environments. Swift verifies
+that all code paths assign a value to the constant at compile time.
+In the example above, it's important that both branches of the `if` statement
+assign a value to `maximumNumberOfLoginAttempts`. If one of the branches did not
+assign a value, that code would raise a compilation error.
 
 You can declare multiple constants or multiple variables on a single line,
 separated by commas:
--- a/TSPL.docc/LanguageGuide/TheBasics.md
+++ b/TSPL.docc/LanguageGuide/TheBasics.md
@@ -93,7 +93,7 @@ Use variables only for storing values that change.
 When you declare a constant or a variable,
 you can give it a value as part of that declaration,
 like the examples above.
-Alternately,
+Alternatively,
 you can assign its initial value later in the program,
 as long as it's guaranteed to have a value
 before the first time you read from it.
--- a/TSPL.docc/LanguageGuide/TheBasics.md
+++ b/TSPL.docc/LanguageGuide/TheBasics.md
@@ -88,7 +88,7 @@ because this value must be incremented after each failed login attempt.
 
 If a stored value in your code won't change,
 always declare it as a constant with the `let` keyword.
-Use variables only for storing values that need to be able to change.
+Use variables only for storing values that change.
 
 When you declare a constant or a variable,
 you can give it a value as part of that declaration,
@@ -135,28 +135,14 @@ it has a value of 100;
 in any other environment, its value is 10.
 Both branches of the `if` statement
 initialize `maximumNumberOfLoginAttempts` with some value,
-which guarantees that the constant always gets a value.
+guaranteeing that the constant always gets a value.
 
 When you define a constant or variable
 without giving it a value,
-Swift analyzes your code at compile time
-and confirms that all possible paths of execution
-are guaranteed to set a value before reading a value.
-This analysis is called *definitive initialization*.
-
-> Note:
-> Definitive initialization
-> can't construct proofs that require domain knowledge,
-> and its ability to track state across conditionals is finite.
-> If you can determine that a value is always set,
-> but Swift can't prove this is the case,
-> try simplifying the conditionals or use a variable instead.
-
-<!--
-In the most general case,
-DI reduces to the halting problem,
-as shown by Rice's theorem.
--->
+Swift analyzes your code to make sure you set a value
+before the first time you read its value.
+For more information about this analysis,
+see <doc:Declarations#Constant-Declaration>.
 
 You can declare multiple constants or multiple variables on a single line,
 separated by commas:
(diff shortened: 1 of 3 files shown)


amartini51:
Thanks for working on this - I think this is a good starting point, and this looks like the right place in the guide to add this information.  After the code listing, I think we need a couple sentences that walk the reader through how the listing works, in addition to telling them what it does.  To say that another way, we should highlight the fact that both branches of the `if` set a value, so there's guaranteed to be a value before any later code tries to read from the constant.  Would you like to try writing that, or should I?

Don't worry too much about the swifttest syntax.  Your markup looks correct, except for one indentation fix I made.  When exporting from ReStructured Text to DocC markdown, we preserved the hidden test code so that some day we could start testing the code listings again, but it isn't actually used for anything right now.

natikgadzhi:
I’m happy to polish this! Down with Covid, though — might take a day or two to get to it.

natikgadzhi:
@amartini51, here's an attempt! 

Thank you for taking a look earlier <3 

>Don't worry too much about the swifttest syntax. Your markup looks correct, except for one indentation fix I made. When exporting from ReStructured Text to DocC markdown, we preserved the hidden test code so that some day we could start testing the code listings again, but it isn't actually used for anything right now.

Ah, I thought it actually works, and I thought it's a really nice way to embed tests into the book and make sure all code examples actually compile and work. I just copied from another example and tried to match the formatting. Thank you for cleaning it up — I'll match indentation rules the next time!

amartini51:
> I’m happy to polish this!

Here's another page we can take examples and content from:

https://developer.apple.com/swift/blog/?id=28

I forgot about this blog post, but found it today when looking for something else.

natikgadzhi:
@amartini51 thank you for the link! 

I took a few days away from this, re-read The Basics page, and the post you linked. I thought about how I can best explain this topic to a newcomer, in the spirit of gradual exposure to complexity, but also provide context to learn more for folks who are curious.

The resulting text links to the Wiki page for definitive initialization, and groups the paragraph on how Swift does this after the example.

Do you think this would be a good way to explain the topic, or am I missing something? How can I improve on this?

natikgadzhi:
@amartini51, friendly ping. How does it look to you now? Does that wording make better sense?

amartini51:
@natikgadzhi Thanks for letting me know this is ready for review. It will be a few more days before I can go through it in detail.  Currently trying to finish some additions to the reference about macros, and moving between office buildings.

amartini51:
@natikgadzhi Thanks for your patience while I was finishing work on the Swift 5.9 version of TSPL.  I've revised the prose in this PR and pushed the changes to your branch.  Please have a look and make any additional changes you think it needs — when you're satisfied with the PR, I'll add an editorial reviewer.

natikgadzhi:
Looks good to me! I think the key thing that you got right is that the change is not just about constant initialization, but about how it makes sense to explain whether to choose a constant or a variable AND how to use them more cohesively. 👏

amartini51 (approved):

TSPL.docc/LanguageGuide/TheBasics.md
```diff
@@ -86,6 +86,72 @@ because the maximum value never changes.
...
+are guaranteed to set a value before reading a value.
+This analysis is called *definitive initialization*.
+
+> Note:
+> Definitive initialization
+> can't construct proofs that require domain knowledge,
+> and its ability to track state across conditionals is finite.
+> If you can determine that a value is always set,
+> but Swift can't prove this is the case,
+> try simplifying the conditionals or use a variable instead.
```
amartini51, line 153: @chuckdude I rewrote this note once, but I'm still not quite happy with it.  Essentially, the full generality of DI reduces to the halting problem, so it's provably impossible for the compiler to always detect.  Maybe this detail should move to the reference.

TSPL.docc/LanguageGuide/TheBasics.md
```diff
@@ -86,6 +86,72 @@ because the maximum value never changes.
...
+Both branches of the `if` statement
+initialize `maximumNumberOfLoginAttempts` with some value,
+which guarantees that the constant always gets a value.
+
+When you define a constant or variable
+without giving it a value,
+Swift analyzes your code at compile time
+and confirms that all possible paths of execution
+are guaranteed to set a value before reading a value.
+This analysis is called *definitive initialization*.
```
amartini51, line 145: @chuckdude Is there a preference from Editorial between calling this "definite initialization" vs "definitive initialization"? The [compiler jargon list](https://github.com/apple/swift/blob/main/docs/Lexicon.md#di-definite-initialization--definitive-initialization) lists both terms. The [blog post on developer.apple.com](https://developer.apple.com/swift/blog/?id=28) uses "definitive" in its link to the Wikipedia article titled "Definite assignment analysis".

I prefer "definite" because it seems more accurate: The compiler allows deferred initialization only when it can prove at compile time that a value is definitely assigned on every code path. There's nothing definitive involved, in the sense of being authoritative or final. The name likely comes from the three states a constant/variable can be in during analysis: definitely assigned (the compiler proved it will have a value set, before being read) definitely unassigned (it proved the opposite) and unknown (it couldn't prove either).

If you agree with "definite", I'll dig into who should review that term from the compiler side and we can add it to a style guide.

chuckdude (approved):
Approved with changes. Let me know if you'd like to revisit the Note separately.

TSPL.docc/LanguageGuide/TheBasics.md
amartini51: Moved DI details to the reference.  Now I'm debating whether an xref is even needed.

TSPL.docc/LanguageGuide/TheBasics.md
```diff
@@ -86,6 +86,78 @@ because the maximum value never changes.
 The current login attempt counter is declared as a variable,
 because this value must be incremented after each failed login attempt.
 
+If a stored value in your code won't change,
+always declare it as a constant with the `let` keyword.
+Use variables only for storing values that need to be able to change.
```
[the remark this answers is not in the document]
amartini51: Changed to "values that change".

bjlanier (approved):

PR #173, opened 2023-08-10
SE-0377: Document the `borrowing` and `consuming` parameter modifiers

tbkka:
This documents the `borrowing` and `consuming` parameter modifiers introduced by [SE-0377][].

These are comparatively niche, so I'm only documenting them in the "Reference Manual" section for now.

I've structured this by:

* Creating a new "Parameter Modifiers" section with a brief description of the modifiers concept,
* Moving the old "In-Out Parameters" description under it, and
* Adding a new subsection on Borrowing and Consuming Parameters.

I've also adjusted the "In-Out Parameters" introduction so it starts with a code example showing the `inout` syntax and usage.

[SE-0377]: https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md

Fixes: rdar://116031853

### DIFF
--- a/TSPL.docc/ReferenceManual/Declarations.md
+++ b/TSPL.docc/ReferenceManual/Declarations.md
@@ -1045,7 +1045,6 @@ see <doc:Functions#In-Out-Parameters>.
 By default, Swift uses a set of rules
 to automatically manage manage object lifetime across function calls,
 copying values when required.
-<!-- XXX "object lifetime" is a new term -->
 The default rules are designed to minimize overhead in most cases ---
 if you want more specific control,
 you can apply the `borrowing` or `consuming` parameter modifier.
@@ -1058,13 +1057,15 @@ ownership are correctly managed in all cases.
 These parameter modifiers impact only the relative efficiency
 of particular usage patterns, not correctness.
 
-<!-- XXX xref the ARC chapter, for info about the default behavior -->
+<!-- XXX TR:
+Should we describe the default rules somewhere?
+Here or in the ARC chapter?
+-->
 
 The `borrowing` modifier indicates that the function
 does not keep the parameter's value.
-In this case, the caller retains ownership
-and is responsible for ensuring that the object remains alive.
-<!-- XXX we don't use "alive" elsewhere in this sense -->
+In this case, the caller maintains ownership of the object
+and the responsibility for the object's lifetime.
 Using `borrowing` minimizes overhead when the function
 is making only transient use of the object.
 
--- a/TSPL.docc/ReferenceManual/Declarations.md
+++ b/TSPL.docc/ReferenceManual/Declarations.md
@@ -1648,20 +1648,15 @@ but the new method must preserve its return type and nonreturning behavior.
 >
 > *parameter-clause* → **`(`** **`)`** | **`(`** *parameter-list* **`)`** \
 > *parameter-list* → *parameter* | *parameter* **`,`** *parameter-list* \
-> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *type-annotation* *default-argument-clause*_?_ \
-> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *type-annotation* \
-> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *type-annotation* **`...`** \
+> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *parameter-type-annotation* *default-argument-clause*_?_ \
+> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *parameter-type-annotation* \
+> *parameter* → *external-parameter-name*_?_ *local-parameter-name* *parameter-type-annotation* **`...`** \
+>
 > *external-parameter-name* → *identifier* \
 > *local-parameter-name* → *identifier* \
-> *default-argument-clause* → **`=`** *expression*
-
-<!-- XXX merge into the box above -->
-
-> Grammar of a function parameter:
->
-> *parameter* → *external-parameter-name*_?_ *local-parameter-name* **`:`** *attributes*_?_ *parameter-modifier*_?_ *type* \
+> *parameter-type-annotation* → **`:`** *attributes*_?_ *parameter-modifier*_?_ *type*
 > *parameter-modifier* → **`inout`** | **`borrowing`** | **`consuming`**
-
+> *default-argument-clause* → **`=`** *expression*
 
 <!--
   NOTE: Code block is optional in the context of a protocol.
--- a/TSPL.docc/ReferenceManual/Declarations.md
+++ b/TSPL.docc/ReferenceManual/Declarations.md
@@ -1043,7 +1043,7 @@ see <doc:Functions#In-Out-Parameters>.
 #### Borrowing and Consuming Parameters
 
 By default, Swift uses a set of rules
-to automatically manage manage object lifetime across function calls,
+to automatically manage object lifetime across function calls,
 copying values when required.
 The default rules are designed to minimize overhead in most cases ---
 if you want more specific control,
--- a/TSPL.docc/ReferenceManual/Declarations.md
+++ b/TSPL.docc/ReferenceManual/Declarations.md
@@ -833,17 +833,16 @@ repeatGreeting("Hello, world!", count: 2) //  count is labeled, greeting is not
 
 ### Parameter Modifiers
 
-A parameter can also include one of the *parameter modifier*
-keywords `inout`, `borrowing`, or `consuming`
-to modify how the
-argument is passed to the function.
+A *parameter modifier* changes how the argument is passed to the function.
 
 ```swift
 <#argument label#> <#parameter name#>: <#parameter modifier#> <#parameter type#>
 ```
 
-You can specify at most one parameter modifier next to the type
-of the argument:
+To use a parameter modifier,
+write `inout`, `borrowing`, or `consuming`
+before the argument's type.
+
 ```swift
 func someFunction(a: inout A, b: consuming B, c: C) { ... }
 ```
@@ -851,8 +850,10 @@ func someFunction(a: inout A, b: consuming B, c: C) { ... }
 #### In-Out Parameters
 
 By default, function arguments in Swift are passed by value:
-any changes made within the function are not visible in the caller.
-You can change this with the `inout` parameter modifier.
+Any changes made within the function are not visible in the caller.
+To make an in-out parameter instead,
+you apply the `inout` parameter modifier.
+
 
 ```swift
 func someFunction(a: inout Int) {
@@ -860,14 +861,14 @@ func someFunction(a: inout Int) {
 }
 ```
 
-When calling such a function, the `inout` argument is
-prefixed with a `&` symbol to mark that the value may
-be changed by the function call.
+When calling a function that includes in-out parameters,
+the in-out argument prefixed with an ampersand (`&`)
+to mark that the function call can change the argument's value.
 
 ```swift
 var x = 7
 someFunction(&x)
-print(x) // 8
+print(x)  // Prints "8"
 ```
 
 In-out parameters are passed as follows:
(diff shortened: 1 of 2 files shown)


TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1044,117 @@ see <doc:Functions#In-Out-Parameters>.
...
+notation when you call the function:
+
+```swift
+func someFunction(a: borrowing A, b: consuming B) { ... }
+
+someFunction(a: someA, b: someB)
+```
+
+The `borrowing` modifier indicates that the function
+does not expect to keep the value.
```
amartini51, line 1067: Let's rephrase this as a more rigorous definition.  As written, it suggests that a function might sometimes keep a borrowed value, possibly "unexpectly" and without copying.

This sounds like it might be a usage guideline — you typically write `borrowing` when a function uses a value only during the function call, without any need to keep the value for later.

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1044,117 @@ see <doc:Functions#In-Out-Parameters>.
...
+Conversely, the `consuming` modifier indicates
+that the function *does* expect to keep the value,
+either to store it or to destroy it.
+This implicitly transfers ownership from the
+caller to the function,
+minimizing overhead when the caller no longer
+needs to use the object after the call.
+But if the caller does need to use the object
+after the call, it might need to make a private
+copy before calling the function.
```
amartini51, line 1094: As with `borrowing`, let's separate the definition from the usage guidelines.

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1044,117 @@ see <doc:Functions#In-Out-Parameters>.
...
+func store(a: consuming A) {
+    someGlobalVariable = a
+}
+```
+
+Note that Swift guarantees that object lifetime and
+ownership will be correctly managed in all cases,
+with or without one of these modifiers.
+These modifiers only impact the relative efficiency
+for particular patterns of usage.
```
amartini51, line 1107: Let's move into the section intro — the reason to use `borrowing` or `consuming` is for efficiency (decrease the need for copying); the code is still correct without them.

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1044,117 @@ see <doc:Functions#In-Out-Parameters>.
...
 -->
 
+#### Borrowing and Consuming Parameters
+
+The Swift compiler has varying ownership rules that it
+can use to manage object lifetimes across function calls.
+The default rules work well for most cases,
+but if necessary, you can use one of the `borrowing`
+or `consuming` modifiers to choose a specific ownership
+rule for a particular argument.
```
amartini51, line 1054: Can we give more specific guidance than "when necessary"?  From your note below, it sounds like this is needed only for performance reasons.

amartini51 (changes requested):
Thanks, Tim!

Structurally, I think this is the right place to add this information, and I think the demotion of the in-out section to make a sibling of the borrowing/consuming section is the right approach.

Some of the content about `borrowing` and `consuming` is written more like part of the guide than reference, but I can help you clean that up later.  If this is feeling too long for the reference, we can start a section in Memory Safety about ownership.

We probably don't need to add examples of exclusivity violations in the reference, because there's already [a section](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/memorysafety/#Conflicting-Access-to-In-Out-Parameters) that discusses these situations in more detail in "Memory Safety", which is linked at the end of the section.

The formal grammar needs to move at the end of sections, and needs to start with a "Grammar of" line.  (When we get grammar we're happy with, we have to update the summary of the grammar manually too.)  The new *parameter-modifier* needs to be added to one of the other productions -- currently, nothing in the grammar refers to this production, so there's no way to "get here" so to speak.  I think it could be added to the definition of *parameter*, although I also see `inout` is already part of `type-annotation`, so that might be the place to add `borrowing` and `consuming`.

A minor point: How did you decide where to break lines?  It looks like you might be hard wrapping at ~60 characters?  The existing content breaks lines at clause boundaries, with the idea that each line is a "chunk" of prose that stands on its own to some extent.  Most of the existing notes in the [TSPL style guide](https://github.com/apple/swift-book/blob/main/Style.md#semantic-line-breaks) about this tell you why to use semantic line breaks, but maybe more detail there is needed about exactly where to break lines.

TSPL.docc/ReferenceManual/Declarations.md
tbkka: Good point.  Changed "does not expect to" to "will not" (which is accurate).

tbkka:
>  I also see inout is already part of type-annotation, so that might be the place to add borrowing and consuming.

Umm....  Is that right?   In particular, I don't think we currently allow `var foo: inout Int` (which is suggested by the grammar here).

tbkka:
>  The new parameter-modifier needs to be added to one of the other productions ...

I was a little confused about that, because the earlier sections in this particular file do not provide grammar productions for function declarations, just template examples.

tbkka:
> The existing content breaks lines at clause boundaries, ...

I believe I've done that.  If you see any place I've missed, please let me know (I'm certainly familiar with the old nroff/manpage conventions and do try to follow them strictly.  But yes, I do tend to prefer short lines.)

TSPL.docc/ReferenceManual/Declarations.md
tbkka: Done.

TSPL.docc/ReferenceManual/Declarations.md
tbkka: Done.

TSPL.docc/ReferenceManual/Declarations.md
tbkka: Clarified a bit.  Let me know if you think this still needs more.

amartini51:
@tbkka

>> I also see inout is already part of type-annotation, so that might be the place to add borrowing and consuming.
>
> Umm.... Is that right? In particular, I don't think we currently allow var foo: inout Int (which is suggested by the grammar here).

Looking back through the commit history, that part of the grammar is from commit ba967673b766a186bbf77f671cf609b99508f04d, and came from the grammar in [SE-0031](https://github.com/apple/swift-evolution/blob/main/proposals/0031-adjusting-inout-declarations.md).  Prior to that commit, `inout` used to be part of the grammar for 'parameter'.  Should we include those proposal authors as reviewers, once you and I have a PR that's ready for final technical and editorial review, or include them in the discussion about this grammar?

>>The new parameter-modifier needs to be added to one of the other productions ...
>
> I was a little confused about that, because the earlier sections in this particular file do not provide grammar productions for function declarations, just template examples.

Maybe it's hard to follow while editing, but the formal grammar appears at the end of each section.  So subsections don't contain their own grammar.  For example, the Variable Declaration section has several subsections, and then one "Grammar of a variable declaration" box at the end.

Each section ends with a grammar-of box — in the case of Function Declaration, there are nine subsections, so you have to scroll a fair bit.  The new grammar productions in all nine of those subsections are collected in the "Grammar of a function declaration" box at the end of the section.

Elsewhere in the reference, some subsections have their own grammar-of boxes.  For function declarations, I think they're all in one box mostly because the subsections don't introduce much grammar — most of the grammar is the same for regular functions as it is for functions that are throwing, async, nonreturning, and so on.  But maybe the Functions section is getting too long, and we should divide up its grammar to make it easier to navigate.

>> The existing content breaks lines at clause boundaries, ...
>
> I believe I've done that. If you see any place I've missed, please let me know (I'm certainly familiar with the old nroff/manpage conventions and do try to follow them strictly. But yes, I do tend to prefer short lines.)

You're exactly correct; the convention here is essentially the same as nroff and man pages.  Different writers have different styles for where to break lines and how long a line should be, which is fine.  (I can often guess who wrote something in TSPL, based on formatting preferences.)

To help keep the content readable in a monospace font, let's prefer ` --- ` rather than an em-dash.  I see one or two places elsewhere in the book where those are used, which I'm fixing.

I'll take a light revision pass for style and, and then push to this branch.

amartini51:
@tbkka Please take a look at my revised version.  Most of the changes are Apple or DevPubs style, with some my personal wording preference, and a couple minor changes to order — introducing things before showing them.  I left XXX markers for things I want to resolve before merging this branch because the build script for TSPL treats them as a hard error, ensuring none get left behind.

I see you have some TODO markers near the end.  Were those things you'd like to include in this PR, or possibilities for the future?

I'm also planning to re-read the SE proposal again while looking at the reference, checking for additional information we should include.

amartini51:
(Ping for @tbkka)

tbkka:
Your changes look good!

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -833,41 +833,42 @@ repeatGreeting("Hello, world!", count: 2) //  count is labeled, greeting is not
...
 func someFunction(a: inout Int) {
     a += 1
 }
 ```
 
-When calling such a function, the `inout` argument is
-prefixed with a `&` symbol to mark that the value may
-be changed by the function call.
+When calling a function that includes in-out parameters,
+the in-out argument prefixed with an ampersand (`&`)
```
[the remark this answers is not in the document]
amartini51: Thanks — fixed in e861c681.

amartini51 (approved):
Before I merge, I want to have someone double check the grammar changes with me.  Otherwise I think this is ready now.

The new to-do markers that this PR adds are both things we can add later:

- `borrowing` and `consuming` keywords with noncopyable argument types.  We don't discuss noncopyable types yet in TSPL (rdar://110848216).

- Any change of parameter modifier is ABI-breaking.  TSPL doesn't discuss ABI stability other than`@frozen`.

jckarter (approved):

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1040,152 @@ see <doc:Functions#In-Out-Parameters>.
...
+store(a: value)
+```
+
+If the caller does need to use the object after the function call,
+it might need to make a separate
+copy before calling the function.
+<!-- XXX TR:
+When do you need to explicitly copy?
+When does the compiler implicitly insert a copy, as shown below?
+-->
```
amartini51, line 1122: @tbkka When does copying need to be explicit, and when is it implicit?

tbkka: I believe that technically, the compiler today always inserts a copy (unless the type is not copyable) and then relies on the optimizer to remove the copy if it turns out to be redundant.  However, I'm not sure we should document this behavior as a requirement of the Swift language.  We should certainly not document it here, as it's really a feature of Swift's general function-calling convention, which is independent of these particular modifiers.

amartini51: Thanks, Tim — removed the XXX marker in e738f1a0, so this doesn't block merging.

Can we rewrite the "might need" bit in the sentence?  As is, it's unclear when you do and don't need to make a copy.

tbkka: Apologies, I think I just realized where we're mis-communicating.  I was thinking of this from the compiler's point of view.  It should probably read:

> If the caller does need to use the object after the function call, the compiler will automatically make a copy of the object if it can.  If the object is non-copyable, the compiler will not be able to do this for you.

amartini51: 👍 Revised in e98ee058

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1040,150 @@ see <doc:Functions#In-Out-Parameters>.
   ```
 -->
 
+#### Borrowing and Consuming Parameters
+
+By default, Swift uses a set of rules
+to automatically manage manage object lifetime across function calls,
```
rexmas, line 1046: managed used twice here, possibly unintentionally

amartini51: Thanks — removed in 728bb4ea.

chuckdude (approved):
Some minor changes throughout; approved with changes.

Please let me know if there's anything here that you'd like to discuss.

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1040,154 @@ see <doc:Functions#In-Out-Parameters>.
...
+```
+
+<!--
+  TODO: `borrowing` and `consuming` keywords with noncopyable argument types
+-->
+<!--
+  TODO: Any change of parameter modifier is ABI-breaking
+-->
+
 ### Special Kinds of Parameters
```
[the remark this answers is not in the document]
amartini51: Stet (but possibly discuss separately).

This is an existing heading, not changed by this PR.

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -980,6 +1040,154 @@ see <doc:Functions#In-Out-Parameters>.
...
+#### Borrowing and Consuming Parameters
+
+By default, Swift uses a set of rules
+to automatically manage manage object lifetime across function calls,
+copying values when required.
+The default rules are designed to minimize overhead in most cases ---
+if you want more specific control,
+you can apply the `borrowing` or `consuming` parameter modifier.
+In this case,
+you must use `copy` to explicitly mark copy operations.
```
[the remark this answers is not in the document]
amartini51: @tbkka Is the "must" needed here to emphasize the action that the developer must take (obligation), or is Chuck's rewrite ok with you?

TSPL.docc/ReferenceManual/Declarations.md
```diff
@@ -865,9 +903,31 @@ Within a function, don't access a value that was passed as an in-out argument,
 even if the original value is available in the current scope.
 Accessing the original is a simultaneous access of the value,
 which violates Swift's memory exclusivity guarantee.
```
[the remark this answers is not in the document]
amartini51: We're in the Swift book; no need to spell this out as being "in Swift".

TSPL.docc/ReferenceManual/Declarations.md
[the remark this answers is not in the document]
tbkka: Either way looks good to me.

bjlanier (approved):

PR #184, opened 2023-09-24
Improve docs for implementing macros in an existing Swift package

ianthetechie:
This contribution improves the documentation for getting started with macros in an existing Swift Package. Notably, it is not obvious to first-time users that `CompilerPluginSupport` needs to be imported, and the default platform support is too low to support macros (in Xcode 15 at the time of this writing).

It also adds imports ahead of the `FourCharacterCode` macro implementation example, since these will be needed to make any(?) macros compile. Tangentially, it looks like #182 addresses the remaining things that tripped me up initially. These combined additions will smooth out the intro to macros significantly.


TSPL.docc/LanguageGuide/Macros.md
```diff
@@ -462,7 +462,25 @@ this creates several files,
 including a template for a macro implementation and declaration.
 
 To add macros to an existing project,
-add a target for the macro implementation
+first ensure that your `swift-tools-version` is set to  5.9 or greater.
+Your `platforms` list must similarly specify a macOS version of 10.15 or greater.
+You also need to import the `CompilerPluginSupport` module.
+The first few lines of your `Package.swift` file might look something like this:
```
amartini51, line 468: I think these three changes are a good candidate for a bulleted list.

```suggestion
Edit your `Package.swift` file as follows:

- Set a Swift tools version of 5.9 or later in the `swift-tools-version` comment
- Import the `CompilerPluginSupport` module.
- Include macOS 10.15 as a minimum deployment target in the `platforms` list

The code below shows the beginning of an example `Package.swift` file.
```

TSPL.docc/LanguageGuide/Macros.md
```diff
@@ -462,7 +462,25 @@ this creates several files,
...
+// swift-tools-version: 5.9
+
+import PackageDescription
+import CompilerPluginSupport
+
+let package = Package(
+    name: "MyPackage",
+    platforms: [
+        .iOS(.v17), .macOS(.v13),
+    ],
```
amartini51, line 480: Let's avoid cutting the example off in the middle of an initializer.  Maybe can show a little more of the file, or an explicit `...` followed by the close paren.

amartini51 (commented):
Thanks for these additions!  I took a pass at rewriting some of the prose.

ianthetechie (commented):
Thanks for the review @amartini51! I've pushed a few changes.

1. Accept your suggestion re: the bulleted list. Good call.
2. I noticed that the use of ending punctuation was inconsistent. I added it, as that seemed to be the norm, but let me know if you'd prefer otherwise.
3. I reorganized the part about adding the SwiftSyntax dependency to be in the list + `Package.swift` sample. That left the remaining original text a bit sparse (another single-sentence paragraph). I'd welcome any style suggestions on that.

TSPL.docc/LanguageGuide/Macros.md
```diff
@@ -462,7 +462,36 @@ this creates several files,
...
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "some-tag"),
+    ],
+    ...
```
amartini51, line 490: Matching how we spell "more code goes here" elsewhere:

```suggestion
    // ...
```

ianthetechie:
Quick bump on this as it's getting a bit stale and now merge conflicts are starting to pop up.

amartini51:
@ianthetechie I'll take another look at this — I started a rewrite pass, but then had to focus on getting TSPL updated for Swift 5.9.2 beta 3.  Thanks for resolving the merge conflict.  I was expecting to see some conflicts, since other PRs also modified some of the same content as this one.

There's a balance in this PR — which I also had to make in the first version — between providing enough information about using Swift Syntax for folks to get started, and not trying to actually teach too much of Swift Syntax here.  I'm also trying to make some adjustments to keep the order of steps linear, with alternate instructions at some points for new/existing code, rather than folding too many steps together under an "existing project" alternative.

A general workflow comment: force-pushing to a branch during PR review makes that review more difficult.  As a reviewer, I can no longer easily see what changed since the last time I looked at the PR.  I'm aware that workflows vary here, so you may have gotten the opposite guidance from others in the past.  (For more details, please see [CONTRIBUTING](https://github.com/apple/swift-book/blob/main/CONTRIBUTING.md) in this repo.)

ianthetechie:
Understood; sorry about that force push!

amartini51:
@ianthetechie How's my latest commit look to you?  Some content that was recently merged into 'main' got deleted as part of merging that branch in here, which I restored.  I also removed some duplicate discussion of `SwiftSyntax`.  My intended approach was to keep the same step-by-step ordering in the content, but to add "if new" and "if existing" versions of the steps where needed.

TSPL.docc/LanguageGuide/Macros.md
```diff
@@ -462,10 +462,32 @@ this creates several files,
...
+    platforms: [ .iOS(.v17), .macOS(.v13)],
+    // ...
+)
+```
+
+Next, add a target for the macro implementation
+and a target for the macro library
+to your existing `Package.swift` file.
 For example,
-you can add something like the following to your `Package.swift` file,
```
amartini51, line 468: Moved the mention of `Package.swift` earlier, to help clarify that this instruction is for existing packages, not newly generated ones.

TSPL.docc/LanguageGuide/Macros.md
```diff
@@ -505,9 +527,6 @@ dependencies: [
 ],
 ```
 
-Replace the `some-tag` placeholder in the code above
-with the Git tag for the version of SwiftSyntax you want to use.
```
amartini51, line 509: The example no longer has a placeholder.  I fixed this on 'main' after you last merged that in here, but then I must have picked up the deletion here while backing out content that was lost by that merge.

ianthetechie: Gotcha; I think I may have merged incorrectly, but I thought I had noticed an inconsistency where both the reference to "some tag" and the raw version number were included and I guess I resolved something wrong.

This makes sense now.

ianthetechie (commented):
Thanks @amartini51! I think this revision is much better, and I appreciate your attention to the flow of things. All looks good to me.

PR #186, opened 2023-09-29
Corrected variable type in 'Optional Binding' section of TheBasics.md.

MasoudIV:
In this section, I noticed a small inaccuracy in the documentation. It mentioned that the type of possibleNumber is Int?, but it should correctly state that the type of Int(possibleNumber) is Int?. I have made the necessary correction for accuracy.


amartini51:
Can you describe the problem in more detail please?  In this text, `possibleNumber` is the name of a variable.  Variables have a type.  The type of the variable `possibleNumber` is `Int?` or "optional Int".

Your change, `Int(possibleNumber)` is not a variable — that code calls an `Int` initializer from, passing `possibleNumber` as an argument.  That code turns an optional Int into a non-optional Int.

I believe this content is correct as-is, but I want to make sure I understand what you were trying to fix.

PR #189, opened 2023-10-03
Fix the first letter of a referencing function name in Macros.

es-kumagai:
In Macro section, a function is referenced as `FourCharacterCode`, but its actual definition is `fourCharacterCode`.
Correcting the reference.

### DIFF
--- a/TSPL.docc/LanguageGuide/Macros.md
+++ b/TSPL.docc/LanguageGuide/Macros.md
@@ -588,7 +588,7 @@ In the example above,
 the first `guard` block extracts the string literal from the AST,
 assigning that AST element to `literalSegment`.
 The second `guard` block
-calls the private `FourCharacterCode(for:)` function.
+calls the private `fourCharacterCode(for:)` function.
 Both of these blocks throw an error if the macro is used incorrectly ---
 the error message becomes a compiler error
 at the malformed call site.


es-kumagai:
Sorry, 4ea92153e7d22fd651863f5506387b8ac039e04d has a wrong commit message.

amartini51 (approved):
Confirmed this is a typo.  The function is defined on line 541.

es-kumagai:
I'd like to fix the commit message,
from: In Macro section, a function is referenced as , but its actual definition is .
to: In Macro section, a function is referenced as 'FourCharacterCode', but its actual definition is 'fourCharacterCode'.

amartini51:
@es-kumagai Do you want to force push your fix to the commit?  Or is it ok if I edit the commit message and force push to your branch?

I think that commit is also ok if you just write "Fix typo in function name".  Because anybody who reads the commit will see in the diff what the typo was.

Thanks for the three fixes today.  If you find more minor errors in the future, you can make one PR with multiple corrections.

es-kumagai:
Thank you for your reply.
That commit message spans multiple lines.
I'd like to fix and force push, I will try it now.

es-kumagai:
I've completed the fixes.

amartini51:
Thanks, I see your change, but now the commit message has very long lines.  This makes the commit hard to read, because many Git tools expect lines <80 characters.  This is a minor issue, but I'll correct it for you.

PR #190, opened 2023-10-03
The type `FourCC` is incorrectly used in the test code of Macros.

es-kumagai:
Fixed a test code referencing a non-existent type `FourCC`. Changed it to `FourCharacterCode`.


amartini51 (approved):
This looks like a copy/paste error during development of the sample.  The type was originally named `FourCC` but expanded to `FourCharacterCode` for readability.

Confirmed that `FourCC` doesn't appear anywhere in the Xcode project that I have on my computer, where I build and test this macro example.

PR #191, opened 2023-10-03
Replace 'SwiftSystem' with 'SwiftSyntax' in Macros section.

es-kumagai:
The term `SwiftSystem` appears in Macros section, but I don't know what it references.
I'm not entirely sure, but from the context, it might be referring to `SwiftSyntax`.

Based on this reasoning, I replaced `SwiftSystem` with `SwiftSyntax`.


amartini51:
Thanks for this correction.  This was a typo — [Swift System](https://github.com/apple/swift-system) is a different module.

amartini51 (approved):
Confirmed that there are no other places that need to change.

Swift System is mentioned in one hidden comment in Concurrency, noting that an example there is a simplified version of a real API.

PR #208, opened 2023-11-14
completing access modifiers example, by also adding open's default members access level.

deya-eldeen:
readers might be in doubt about the `open` access modifier, especially `internal` members, adding this part will explain this more and remove doubts.

we can remove some parts of it for brevity, but this line is important

`    var someInternalProperty = 0                 // implicitly internal class member`

### DIFF
--- a/TSPL.docc/LanguageGuide/AccessControl.md
+++ b/TSPL.docc/LanguageGuide/AccessControl.md
@@ -231,6 +231,14 @@ the default access level of the type's members will be internal.
 > and avoids presenting the internal workings of a type as public API by mistake.
 
 ```swift
+open class SomeOpenClass {                      // explicitly open class
+    open var someOpenProperty = 0                // explicitly open class member
+    public var somePublicProperty = 0            // explicitly public class member
+    var someInternalProperty = 0                 // implicitly internal class member
+    fileprivate func someFilePrivateMethod() {}  // explicitly file-private class member
+    private func somePrivateMethod() {}          // explicitly private class member
+}
+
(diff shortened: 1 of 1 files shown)


amartini51:
Thanks for starting this discussion!  A couple issues that make this change a bit challenging to review:

- The commit message should describe what you changed and why, not just restate which file contains the changes.  (If you're editing on GitHub.com you need to replace the default message it gives you.)
- The pull request should describe the problem in the current text.  In this case, what are developers confused about?  What incorrect conclusion are they likely to arrive at?
- The pull request should describe why this solution resolves that problem.  For example, here, why is it sufficient to add examples to the code?  What about the prose nearby — does that need updates?

For more detailed guidance, please see [CONTRIBUTING](https://github.com/apple/swift-book/blob/main/CONTRIBUTING.md) in this repository.

amartini51:
@deya-eldeen Is this a change you're still interested in making or advocating for?  Or can I close this PR as "not to be merged"?

deya-eldeen:
Still interested, I will provide details about this next week, hope this is
ok.

Deya Eldeen Elkhawaldeh

amartini51:
@deya-eldeen One year later, still interested?  If so, please focus on explaining what problem this change solves in the book.  Even if you don't have a complete solution, understanding the problem helps me (and others) work towards fixing that issue.  Thanks!

To answer one question ahead of time, this code listing isn't intended to be an exhaustive list of every possible access control level.  For example, it omits `open` as you discussed but also `package`.

PR #210, opened 2023-12-02
Fix a few typos

Blake-Madden:
Fix a few typos

### DIFF
--- a/Style.md
+++ b/Style.md
@@ -149,7 +149,7 @@ See entry for *function*.
 
 ## non-optional
 
-Hyphenated to avoid the misreading as nono-ptional.
+Hyphenated to avoid the misreading as non-optional.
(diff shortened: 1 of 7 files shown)


Style.md
```diff
@@ -149,7 +149,7 @@ See entry for *function*.
 
 ## non-optional
 
-Hyphenated to avoid the misreading as nono-ptional.
+Hyphenated to avoid the misreading as non-optional.
```
amartini51, line 152: This is actually correct as is — it's calling out how "nonoptional" can be misread without the hyphen.

```suggestion
Hyphenated to avoid the misreading as nono-ptional.
```

amartini51 (approved):
Thanks for the typo fixes!  All look good, except for one that I called out.

PR #223, opened 2023-12-24
Fix grammar error in The Basics

urcades:
In the sentence contained between lines 1612 and 1615 of "The Basics", a grammatical error before the word "programmer" can be found:

> When nil represents an unrecoverable failure, such **_a_** programmer error or corrupted state, you can access the underlying value by adding an exclamation mark (!) to the end of the optional’s name.

By adding the word "as" before "a", the sentence becomes grammatically correct:

> When nil represents an unrecoverable failure, such **_as a_** programmer error or corrupted state, you can access the underlying value by adding an exclamation mark (!) to the end of the optional’s name.


amartini51 (approved):

amartini51:
Thanks for the correction!

PR #225, opened 2023-12-28
Update OS and arch grammar according the open source Swift repo

Kyle-Ye:
Context:

I was adding some `#if os(xx)` to detect the pointer size on my package, and they failed to build on watchOS platform. I realized I have missed some arch for it. (eg. `armv7k` and `arm64_32`)

I checked the grammar page on TSPL. They are both not supported according to TSPL grammar here. And actually `arm64_32` is supported and `armv7k` is not which will hit `os(arm)` here.

This PR updates the corresponding grammar rule according to https://github.com/apple/swift

I did not find the exact grammar parsing logic so there may be some missing item. But I have checked what I list here are all supported on Swift 5.9

visionOS is not added on upstream open source Swift yet. And they are only supported on Xcode's Swift toolchain. (eg. Xcode 14.0)

Building the following code on Linux with the latest Swift release(Swift 5.9.2) will give us a warning here.
```
#if os(visionOS)
...
#endif
```

```
DemoKit.swift:12:8: note: did you mean 'iOS'?
#if os(visionOS)
       ^~~~~~~~
       iOS
```

For Arch, I use the following source
- https://github.com/apple/swift/blob/260cec969490db027fa2392c4061c2fe1eba22f6/lib/Basic/LangOptions.cpp#L69-L81
- https://github.com/apple/swift/blob/260cec969490db027fa2392c4061c2fe1eba22f6/stdlib/public/core/AtomicInt.swift.gyb#L85-L91

For OS, I use the following source
- https://github.com/apple/swift/blob/260cec969490db027fa2392c4061c2fe1eba22f6/lib/Basic/LangOptions.cpp#L52-L67
- https://github.com/apple/swift/blob/260cec969490db027fa2392c4061c2fe1eba22f6/stdlib/public/Platform/Platform.swift#L482

### DIFF
--- a/TSPL.docc/ReferenceManual/Statements.md
+++ b/TSPL.docc/ReferenceManual/Statements.md
@@ -961,8 +961,8 @@ conditions listed in the table below.
 
 | Platform condition | Valid arguments |
 | ------------------ | --------------- |
-| `os()` | `macOS`, `iOS`, `watchOS`, `tvOS`, `visionOS`, `Linux`, `Windows` |
-| `arch()` | `i386`, `x86_64`, `arm`, `arm64` |
+| `os()` | `OSX`, `macOS`, `tvOS`, `watchOS`, `iOS`, `Linux`, `FreeBSD`, `OpenBSD`, `Windows`, `Android`, `PS4`, `Cygwin`, `Haiku`, `WASI` |
+| `arch()` | `arm`, `arm64`, `arm64_32`, `i386`, `x86_64`, `powerpc`, `powerpc64`, `powerpc64le`, `s390x`, `wasm32`, `riscv64` |
 | `swift()` | `>=` or `<` followed by a version number |
 | `compiler()` | `>=` or `<` followed by a version number |
 | `canImport()` | A module name |
@@ -1210,8 +1210,8 @@ see <doc:Expressions#Explicit-Member-Expression>.
 > *platform-condition* → **`canImport`** **`(`** *import-path* **`)`** \
 > *platform-condition* → **`targetEnvironment`** **`(`** *environment* **`)`**
 >
-> *operating-system* → **`macOS`** | **`iOS`** | **`watchOS`** | **`tvOS`** | **`Linux`** | **`Windows`** \
-> *architecture* → **`i386`** | **`x86_64`** | **`arm`** | **`arm64`** \
+> *operating-system* → **`OSX`** | **`macOS`** | **`tvOS`** | **`watchOS`** | **`iOS`** | **`Linux`** | **`FreeBSD`** | **`OpenBSD`** | **`Windows`** | **`Android`** | **`PS4`** | **`Cygwin`** | **`Haiku`** | **`WASI`** \
+> *architecture* → **`arm`** | **`arm64`** | **`arm64_32`** | **`i386`** | **`x86_64`** | **`powerpc`** | **`powerpc64`** | **`powerpc64le`** | **`s390x`** | **`wasm32`** | **`riscv64`** \
 > *swift-version* → *decimal-digits* *swift-version-continuation*_?_ \
 > *swift-version-continuation* → **`.`** *decimal-digits* *swift-version-continuation*_?_ \
 > *environment* → **`simulator`** | **`macCatalyst`**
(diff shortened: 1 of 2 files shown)


Kyle-Ye:
What's your idea on this? cc @amartini51

TSPL.docc/ReferenceManual/Statements.md
```diff
@@ -958,11 +958,11 @@ evaluates to `true` at compile time.
 The *compilation condition* can include the `true` and `false` Boolean literals,
 an identifier used with the `-D` command line flag, or any of the platform
 conditions listed in the table below.
-
+**`OSX`** | **`macOS`** | **`tvOS`** | **`watchOS`** | **`iOS`** | **`Linux`** | **`FreeBSD`** | **`OpenBSD`** | **`Windows`** | **`Android`** | **`PS4`** | **`Cygwin`** | **`Haiku`** | **`WASI`** \
```
krilnon, line 961:
```suggestion

```

Looks like you might have duplicated this line accidentally.

Kyle-Ye: Yes. Thanks for pointing out. Fixed

amartini51:
I don't think we should make this change.  As noted in the in-source comment below this grammar block:

> Some of the OSes and architectures are listed [in the compiler's source] because there's experimental work to port Swift to them.  We won't list them here until they're officially supported.

Are there officially supported OSes or architectures that are missing from TSPL's list?

It might be ok to add `OSX` which is a compatibility alias for `macOS`, but that probably needs some discussion.

As an aside: This PR would be easier to review if you added things first, and then re-ordered them in a separate commit.

Kyle-Ye:
## Answers

> Are there officially supported OSes or architectures that are missing from TSPL's list?

I can see `arm64_32` should be added as they are part of watchOS SDK.

> As an aside: This PR would be easier to review if you added things first, and then re-ordered them in a separate commit.

Got it. The order is kept as how they appear in the source repository.

## Questions

Here is my questions:

1. When people see a build failure on watchOS(due to the missing arch check), what is the expected site for people to check the arch grammer?

IMO They should either go to Apple's documentation or TSPL to check for all the possible arches Swift support to update their code instead of digging into the Swift(Swift Compiler) repo.

https://github.com/OpenCombine/OpenCombine/blob/1c6f02c7ed8140c0ba7a783aaddb6e0685a0037b/Tests/OpenCombineTests/DispatchTests/DispatchQueueSchedulerTests.swift#L245-L259

2. We have content (eg. visionOS) listed in TSPL as if they are supported Swift grammar of 5.9.2. But they are actually only accepted by Apple's private Swift fork / Xcode bundled one instead of the open source one hosted on [apple/swift](https://github.com/apple/swift)

> Building the following code on Linux with the latest Swift release(Swift 5.9.2) will give us a warning here.
> ```
> #if os(visionOS)
> ...
> #endif
> ```
> 
> ```
> DemoKit.swift:12:8: note: did you mean 'iOS'?
> #if os(visionOS)
>        ^~~~~~~~
>        iOS
> ```

I can't find anything string about visionOS here in [apple/swift](https://github.com/apple/swift). I guess it is reasonable since it is still beta and no release version of Xcode has visionSDK too.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/summaryofthegrammar#Statements

image

I think the content in TSPL should be aligned with [apple/swift](https://github.com/apple/swift). If Apple has not merged such patch, we should not present it here too.

amartini51 (changes requested):
Only OSes and architectures that are officially supported should be listed.

amartini51:
I'm getting confirmation on whether arm64_32 is officially supported.  If it is, we should add it here.

OSX was intentionally removed from the list in 1f4cffcfd4c7d9edc6f1d215117b9629ceb0c786

Kyle-Ye:
Update: 

1. visionOS issue was clarified here https://github.com/apple/swift-book/pull/227#issuecomment-1879090576

> Only OSes and architectures that are officially supported should be listed.

2.

For OSes: Only Apple's OS, Linux and Windows are officially supported. So I'll remove the others.

For architectures, `i386`, `x86_64`, `arm` and `arm64` are officially supported. But if we only use those arch to make `#if arch(xx)` logic, we'll hit a uncovered case for watchOS platform. The missing one is `arm64_32` and @amartini51 is confirming with it. Other arches should be removed since they are never officially supported.

amartini51:
Initial feedback is that we shouldn't add `arm64_32` and maybe we shouldn't even have this list in TSPL.  I'll close this PR for now, and I've filed a GitHub issue to track this.

https://github.com/apple/swift-book/issues/245

PR #249, opened 2024-01-19
GuidedTour: Fixed while and repeat-while experiment description

ChristophKaser:
When changing the condition to `m <  0`, it is false from the start, not true from the start.


amartini51 (approved):

amartini51:
Thanks for the correction!

PR #266, opened 2024-02-04
Fix spelling of Fahrenheit in example method

stzn:
Fixed typo (convertFarenheitToCelsius -> convertFahrenheitToCelsius)

As far as I confirmed, there is no duplicated PR.


amartini51 (approved):
Confirmed that this is the correct spelling.

Confirmed via `git grep Farenheit` that this is the only place the misspelling appears.

amartini51:
@stzn Thank you!

PR #270, opened 2024-02-09
Make spelling of "canceled" match stdlib

mattmassicotte:
I have a feeling this may be contentious. However, I noticed that the concurrency section using the spelling "canceled" but the Swift standard library uses "cancelled". I'm not entirely sure which is right, but it seems like the book should match, don't you think?

### DIFF
--- a/TSPL.docc/LanguageGuide/Concurrency.md
+++ b/TSPL.docc/LanguageGuide/Concurrency.md
@@ -587,8 +587,8 @@ The explicit parent-child relationships between tasks has several advantages:
 - When setting a higher priority on a child task,
   the parent task's priority is automatically escalated.
 
-- When a parent task is canceled,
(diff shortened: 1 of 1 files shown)


amartini51:
This is a known incompatibility between style guides.  Apple Style Guide, which this book follows, calls for the spelling "canceled":

> Use one l for the verb cancel—for example canceled, canceling. Use two l’s for the noun cancellation.

https://support.apple.com/guide/applestyleguide/c-apsgb744e4a3/1.0/web/1.0

For historical reasons dating back to NeXTSTEP (in the 1980s), APIs in Foundation use the spelling "cancelled" — and I assume the standard library is following that spelling, for consistency between APIs.

mattmassicotte:
Ah ha - thank you for explaining that! It certainly did seem intentional, so I should have known.

PR #279, opened 2024-02-29
Fixes task cancellation sample code

Diggory:
The code listing tried to access the `isCancelled`  property without specifying `Task.isCancelled`.  That would work in the scope of a task, because there's an instance property with that name.  But it doesn't work in the context of calling `addTaskUnlessCancelled(priority:operation:)` to adding a child task to a task group — in that case, you need to use the type property.


amartini51:
@Diggory Thanks!  I made some additional changes on your branch while reviewing.  Do you have any additional comments before I merge?

Diggory:
Looks good to me. P.s. thanks for your work on the Swift Book.

bjlanier (approved):

amartini51 (approved):

PR #297, opened 2024-04-24
Add missing "Protocol" to the title

stzn:
In the content, "boxed protocol types" is used, but it's not used in the title.
I'm not sure if it's intended or not, but it looks missing "Protocol".

### DIFF
--- a/TSPL.docc/LanguageGuide/OpaqueTypes.md
+++ b/TSPL.docc/LanguageGuide/OpaqueTypes.md
@@ -1,4 +1,4 @@
-# Opaque and Boxed Types
+# Opaque and Boxed Protocol Types
 
 Hide implementation details about a value's type.
 


amartini51:
Thanks for this PR!

I also discussed this change with @chuckdude and @bjlanier.  I wasn't sure if naming the chapter "Opaque and Boxed Protocol Types" might be misread as "opaque and boxed" modifying "protocol types".  Spelling it out fully as "Opaque Types and Boxed Protocol Types" is a little too long.  They agreed that this change makes sense, and isn't too easy to misread.

They also noted that we want to make sure the URL doesn't change.  I'm pretty sure the URL comes from the filename, not the title — for example, A Swift Tour is at <https://docs.swift.org/swift-book/documentation/the-swift-programming-language/guidedtour> — but I will confirm when building.

amartini51 (approved):

amartini51:
Confirmed the URL path remains `/opaquetypes` for the chapter we're changing.  Spot-checked the diff in the build product between this PR and 'main' — no unexpected changes to the JSON.  Looks good!

PR #301, opened 2024-05-05
Spell file-private with hyphen

SimplyDanny:
The keyword `fileprivate` referred to in text is spelled in different ways throughout the book. "file-private" with a hyphen appears to me as the preferred style.

### DIFF
--- a/TSPL.docc/LanguageGuide/AccessControl.md
+++ b/TSPL.docc/LanguageGuide/AccessControl.md
@@ -137,7 +137,7 @@ the code in your app is typically self-contained within the app
 and doesn't need to be made available outside of the app's module.
 The default access level of internal already matches this requirement.
 Therefore, you don't need to specify a custom access level.
-You may, however, want to mark some parts of your code as file private or private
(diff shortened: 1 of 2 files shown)


krilnon (changes requested):
I think these are correct as-is. When we spell it "file-private" with the hyphen, it's being used as part of a compound adjective. Otherwise it's just "file private".

SimplyDanny:
> I think these are correct as-is. When we spell it "file-private" with the hyphen, it's being used as part of a compound adjective. Otherwise it's just "file private".

But isn't it always used as an adjective in these cases?

amartini51:
Thanks for taking the time to look at these in the book.  I agree with Kyle — I think these follow our style the way they are.

We hyphenate when "file private" modifies the word that comes after it, like "a file-private function".

SimplyDanny:
Okay, got the difference now. Compound adjectives preceding a noun are with hyphen, while standalone, they are written without one.

At least so I learned something new ...

PR #302, opened 2024-05-05
Remove redundant sentence

SimplyDanny:
Seems like the message of the previous sentence is pretty much the same.

### DIFF
--- a/TSPL.docc/ReferenceManual/Declarations.md
+++ b/TSPL.docc/ReferenceManual/Declarations.md
@@ -3905,13 +3905,13 @@ Access control is discussed in detail in <doc:AccessControl>.
   Apply this modifier to a declaration to indicate the declaration can be accessed
   only by code within the declaration's immediate enclosing scope.
 
-For the purpose of access control,
-extensions to the same type that are in the same file
-share an access-control scope.
-Private members declared in the type's declaration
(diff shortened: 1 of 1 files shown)


amartini51 (changes requested):
I don't think this is redundant — but it looks like it's unclear:

- If there are multiple extensions in the same file, and those extensions all extend the same type, then all of those extensions have the same access control level.  This file might not be the same as where the extended type is defined.
- If the extensions are in the same file as the type they extend, the extensions have the same access control as the type they extend.

If you'd like to try rewriting for clarity, please go ahead on this PR.  Otherwise, you can let me know and I'll try rewriting.

SimplyDanny:
> If you'd like to try rewriting for clarity, please go ahead on this PR. Otherwise, you can let me know and I'll try rewriting.

I tried to reword the paragraph.

amartini51:
Thanks @SimplyDanny — I think we're going to come back to this.  The rewrite has several issues, and I'm currently focusing on new content for topics like safe concurrency and noncopyable types.  I set a reminder to look at this again in June, when hopefully more time will be available.

amartini51:
Closing this PR in favor of the rewrite on https://github.com/apple/swift-book/pull/315

Thanks for pointing out the hard to read wording!

PR #305, opened 2024-05-15
Fix typo in Methods.md

feelingsonice:
Fix typo

Fixes: rdar://####

### DIFF
--- a/TSPL.docc/LanguageGuide/Methods.md
+++ b/TSPL.docc/LanguageGuide/Methods.md
@@ -410,7 +410,7 @@ Here's how you call a type method on a class called `SomeClass`:
 
 ```swift
 class SomeClass {
-    class func someTypeMethod() {
+    static func someTypeMethod() {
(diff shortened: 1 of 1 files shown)


amartini51:
Type methods on classes use either `static` or `class` depending on whether subclasses are allowed to override the method.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/methods#Type-Methods

> Instance methods, as described above, are methods that you call on an instance of a particular type. You can also define methods that are called on the type itself. These kinds of methods are called type methods. You indicate type methods by writing the static keyword before the method’s func keyword. Classes can use the class keyword instead, to allow subclasses to override the superclass’s implementation of that method.

Why should this method disallow subclasses from overriding it?

feelingsonice:
Ah I miss read the paragraph before. I apologize

amartini51:
Thanks for helping improve the documentation!

Issue #310, opened 2024-06-09
Xcode version is missed


stzn:
### Location

https://github.com/apple/swift-book/blob/main/TSPL.docc/GuidedTour/Compatibility.md#version-compatibility

### Description

This book describes Swift 6, the default version of Swift that's included in Xcode XXX. 

I think `Xcode XXX` should be `Xcode 16`.

### Correction

This book describes Swift 6, the default version of Swift that's included in Xcode 16

amartini51:
This is expected — when an Xcode release hasn't shipped, TSPL doesn't include its version number.  The publication script replaces the XXX here and in the revision history with the actual version number and ship date.  Thanks!

amartini51:
The commits from the publication process that replace the XXX placeholders for Swift 6 beta 1 are:

* 26bb01cf Fill in Swift 6 beta 1 date
* 21f83879 Fill in Xcode version number

PR #311, opened 2024-06-09
Update Xcode version

stzn:
Xcode version is missed. I think Xcode XXX should be Xcode 16.

Fixes: https://github.com/apple/swift-book/issues/310


amartini51:
This placeholder is intentional — see discussion on #310.  The build script I use to publish TSPL replaces in the placeholder here and the placeholder date in the revision history.

PR #316, opened 2024-06-13
Fix typo in macro name

Alc-Alc:
Code
```swift
@attached(member, names: named(RawValue), named(rawValue),
        named(`init`), arbitrary)
```
Prose
```
the `@attached(member)` macro includes arguments after the `named:` label
```

In the code the label is referred to as "`names`", but the prose refers to it as "`named`". This PR suggests a doc fix that uses the correct label ("`names`") in the prose.


amartini51 (approved):
Looks good - thank you!  Confirmed that SE-0389 spells it "names".

PR #320, opened 2024-07-02
Emphasize type erasure in  buildLimitedAvailability

ApolloZhu:
A common misconception that people have about `buildLimitedAvailability` is that it enables the following syntax in result builders:

```swift
@DrawingBuilder var x: Drawable {
    if #available(macOS 99, *) {
        FutureText("Inside.future")
    } else {
        Text("Inside.present")
    }
}
```

and would implement `buildLimitedAvailability` as follows:

```swift
static func buildLimitedAvailability(component: some Drawable) -> some Drawable {
    component
}
```

However, the above code will still result in a crash in an earlier OS where `FutureText` is not available. 

**This change calls out the importance and need to perform type erasure in `buildLimitedAvailability`.**

### DIFF
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -1640,11 +1640,9 @@ The additional result-building methods are as follows:
   or to perform other postprocessing on a result before returning it.
 
 - term `static func buildLimitedAvailability(_ component: Component) -> Component`:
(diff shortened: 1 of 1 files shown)


amartini51:
Per Apollo, suggestion added Doug, Holly, or Pavel as a technical reviewer.

DougGregor (approved):

amartini51 (approved):

Issue #321, opened 2024-07-07
Padding between social icons in the footer


shubhsaur:
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/thebasics

### Description

Twitter and feed icon in the footer of the documentation are coupled together without any gap or padding.

<img width="87" alt="Screenshot 2024-07-06 at 5 06 39 PM" src="https://github.com/apple/swift-book/assets/28146789/438523bb-0567-4af1-bb40-fbcf438882eb">

### Correction

There should be some padding in between the twitter and feed icon in the footer.

<img width="245" alt="Screenshot 2024-07-07 at 6 05 55 PM" src="https://github.com/apple/swift-book/assets/28146789/311413cd-8517-4abd-9092-ac785139f8c1">

PR #322, opened 2024-07-07
fixed padding between footer social icons

shubhsaur:
Fix: #321

### DIFF
--- a/TSPL.docc/footer.html
+++ b/TSPL.docc/footer.html
@@ -41,7 +41,7 @@ See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
           <span class="color-scheme-toggle-label">Auto</span>
         </label>
       </form>
-      <aside>
+      <aside class="footer-social-icons">
       <a href="/atom.xml" title="Subscribe to Site Updates"><i class="feed"></i></a>
       <a href="https://twitter.com/swiftlang" rel="nofollow" title="Follow @SwiftLang on Twitter"><i class="twitter"></i></a>
     </aside>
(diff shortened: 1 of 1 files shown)


TSPL.docc/footer.html
```diff
@@ -110,6 +110,13 @@
         padding: 0;
         border:0
     }
+
+    .footer-social-icons {
+      display: flex;
+      flex-direction: row-reverse;
+      justify-content: center;
```
krilnon, line 117:
```suggestion
      justify-content: right;
```

Right justification would better match how these icons appear on the main swift.org footer.

krilnon (changes requested):
I left a small suggestion for style. 

More broadly, although I think this change would indeed fix the issue for the book's footer, a decent amount of the style info comes from `https://swift.org/assets/stylesheets/docc/footer.css`, which we share with the DocC documentation. It lives over in the [swiftlang/swift-org-website](https://github.com/swiftlang/swift-org-website/tree/main/assets/stylesheets/docc) repository.

Maybe that's a better place to make a change to fix this, since it'd help the DocC docs too.

shubhsaur:
Hey @krilnon, thanks for the suggestion, will look into that way as well, and for the time being have made the requested change

krilnon:
@amartini51 If you have thoughts one way or the other on accepting the quick fix here or waiting for a possible upstream fix, your input is welcome.

amartini51:
Thanks for working on this fix!  (I was out of the office for 2 weeks when this PR first came through.  Thanks for your patience waiting for review.)  I'd prefer to correct the CSS in the main Swift.org repo, and then match that correction here.  I think that will also get you better review on the CSS, since there are more folks working on that repository with an expertise in CSS.  I only know CSS at a general level, and have not kept up with all the latest changes.

amartini51:
Moving this PR to "draft" while the changes are being proposed and reviewed for the main Swift.org website.  After we merge the changes there, we'll make the corresponding changes here.

PR #326, opened 2024-08-07
Distributed module documentation

ktoso:
This is a very early version of `Distributed` module documentation; It is not polished up, may have stray sentences and sections etc.

We'll figure out how to best present the information through the pitch posted on the forums over here: 

I'll also work on improving tone and voice still, right now it is not the most consistent. 

For now, we can discuss what needs to be explained in this documentation and I started a forums thread for this purpose: [TSPL Pitch: Distributed](https://forums.swift.org/t/tspl-pitch-distributed/73756).

### DIFF
--- a/TSPL.docc/LanguageGuide/Distributed.md
+++ b/TSPL.docc/LanguageGuide/Distributed.md
@@ -29,7 +29,7 @@ and only `distributed` methods may be called on it.
 
 ## Thinking in Distributed Actors
 
-In order to build distributed systems successfully you will need to get into the right mindset.
+Building distributed systems is a complex task that requires careful consideration of reliability, maintainability, and scalability. There are various approaches to build such systems, and Swift, as a general-purpose language, offers tools for this, such as actors. However, in order to build distributed systems successfully with this tools, you will need to get into the right mindset.
 
 You can use distributed actors, to break up your program into isolated, independent
 pieces of code, which may be located on different processes or hosts at runtime. This extends the notion of isolation
@@ -343,7 +343,7 @@ func testLocal(score: Score,
 As we can see, an cross-actor call to an actor function caused an implicit `async` effect,
 while the same type of call to a `distributed func` caused an additional `throws` effect that needed to be handled
 with a `try`. Since the method `increment(by:)` itself is not declared throwing by itself, we know that the only
-way this method can fail, is by the underlying transport mechanism failing in sme way.
+way this method can fail, is by the underlying transport mechanism failing in some way.
 
 > Note: Exact transport semantics can vary between actor system implementations, so you should consult the 
 > documentation of the `DistributedActorSystem` your actor is using to get a complete picture of its failure
@@ -585,15 +585,15 @@ actor. In local-only actors, a common way of modeling this is passing a closure 
 In distributed actors, the same can be achieved using distributed method calls rather than closures, like this:
 
 ```swift
-protocol InfoListener: Codable {
-  func additionalInfo(_ info: String) async throws
+protocol InfoListener: DistributedActor, Codable {
+  distributed func additionalInfo(_ info: String)
 }
 
 distributed actor Alice {
-  distributed func call(later: InfoListener) async -> String {
+  distributed func call(later: some InfoListener) async -> String {
     Task.detached {
       // do some asynchronous processing AFTER returning from the 'call' method...
-      later("Here's the info you asked for!")
+      try await later.additionalInfo("Here's the info you asked for!")
     }
 
     return "Thanks for your call!"
@@ -616,7 +616,7 @@ distributed actor Bob {
 
 extension Bob: InfoListener {
   distributed func additionalInfo(_ info: String) {
-    print("later: \(laterReply)")
+    print("later: \(info)")
   }
 }
 
@@ -633,7 +633,8 @@ like to associate the reply with, like this:
 ```swift
 distributed actor Bob {
   func test(alice: Alice) async {
-    let immediateReply = try await alice.call(self)
+    // We have to make sure, in some way, that the `DistributedCallback` instance stays alive to receive the callback
+    let immediateReply = try await alice.call(later: DistributedCallback(name: "Bob", actorSystem: self.actorSystem))
 
     print("immediately: \(immediateReply)")
   }
--- a/TSPL.docc/LanguageGuide/Distributed.md
+++ b/TSPL.docc/LanguageGuide/Distributed.md
@@ -4,10 +4,12 @@ Distributed computing with actors.
 
 Swift has built-in support for building distributed systems using the actor model.
 
-The *Distributed* module extends Swift's "local" actor model, providing
-the ability to natively express that a piece of code may not be located in
-the same process as the calling code. This is expressed using the `distributed`
-keyword, which can be used with actor and function declarations.
+The *Distributed* module extends Swift's "local" actor model,
+providing the ability to natively express
+that a piece of code may not be located in
+the same process as the calling code.
+This is expressed using the `distributed` keyword,
+which can be used with actor and function declarations.
 
 Similar to actors, they make use of actor isolation,
 however unlike them,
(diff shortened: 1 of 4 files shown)


TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+This example models a situation where we call the `Alice` actor, and receive a reply immediately, but at the same time
+we registered a closure to be called at some later point in time when some additional information was looked up by the
+actor. In local-only actors, a common way of modeling this is passing a closure to the actor.
+
+In distributed actors, the same can be achieved using distributed method calls rather than closures, like this:
+
+```swift
+protocol InfoListener: Codable {
```
martialln, line 588:
```suggestion
protocol InfoListener: DistributedActor, Codable {
```

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+  func additionalInfo(_ info: String) async throws
```
martialln, line 589:
```suggestion
  distributed func additionalInfo(_ info: String)
```

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+}
+
+distributed actor Alice {
+  distributed func call(later: InfoListener) async -> String {
```
martialln, line 593:
```suggestion
  distributed func call(later: some InfoListener) async -> String {
```

In this case Later concrete type must be known on Alice side.

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+  }
+}
+```
+
+So instead of a closure, the `Alice` distributed actor is now accepting an `InfoListener`,
+which we'll implement using a distributed actor on the calling side by having Bob be a distributed actor and conform to
+this protocol:
+
+```swift
+distributed actor Bob {
```
martialln, line 609:
```suggestion
distributed actor Bob: Codable {
```

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+// OUTPUT: 
+// immediately: Thanks for your call!
+// later: Here's the info you asked for!
+```
+
+This example models a situation where we call the `Alice` actor, and receive a reply immediately, but at the same time
+we registered a closure to be called at some later point in time when some additional information was looked up by the
+actor. In local-only actors, a common way of modeling this is passing a closure to the actor.
+
+In distributed actors, the same can be achieved using distributed method calls rather than closures, like this:
```
martialln, line 585: We should explain that Alice and Bob must share the same ActorSystem. Also there are two cases: is Bob type known on Alice side or not

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+  func test(alice: Alice) async {
+    let immediateReply = try await alice.call(self)
+
+    print("immediately: \(immediateReply)")
+  }
+}
+
+extension Bob: InfoListener {
+  distributed func additionalInfo(_ info: String) {
+    print("later: \(laterReply)")
```
martialln, line 619:
```suggestion
    print("later: \(info)")
```

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+  }
+}
+
```
martialln, line 622: Using resolvable protocol we could have something like that:

```swift
@Resolvable
protocol InfoListener: DistributedActor, Codable {
    distributed func additionalInfo(_ info: String)
}

distributed actor Alice {
    distributed func call(later: $InfoListener) async -> String {
        Task.detached {
            // do some asynchronous processing AFTER returning from the 'call' method...
            try await later.additionalInfo("Here's the info you asked for!")
        }
        
        return "Thanks for your call!"
    }
}

distributed actor Bob: DefaultDistributedActorSystem.SerializationRequirement {
    func test(alice: Alice) async throws {
        let infoListener = try $InfoListener.resolve(id: self.id, using: self.actorSystem)
        let immediateReply = try await alice.call(later: infoListener)
        
        print("immediately: \(immediateReply)")
    }
}

extension Bob: InfoListener {
    distributed func additionalInfo(_ info: String) {
        print("later: \(info)")
    }
}
```

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+You may also create an per-request actor, rather than having Bob conform to this protocol. Using a per-call actor is
+equivalent to using the closure callback from the initial example, because we can capture any additional state we'd
+like to associate the reply with, like this:
+
+```swift
+distributed actor Bob {
+  func test(alice: Alice) async {
+    let immediateReply = try await alice.call(self)
```
martialln, line 636:
```suggestion
    // We have to make sure, in some way, that the `DistributedCallback` instance stays alive to receive the callback
    let immediateReply = try await alice.call(later: DistributedCallback(name: "Bob", actorSystem: self.actorSystem))
```

martialln (commented):
Mainly puts comments on the last section around Emulating callbacks as I found some errors in code examples

akbashev (commented):
Overall like the page, gives a quick recap of the module! 

Want to reread it a bit later with a fresh mind, but at the moment have two small comments. Added them as `suggestions`, but take them with a grain of salt. 
1. Overall I think would be nice to add some small sentence on why exactly one should use `distributed` module. 
2. Think memory management could also be highlighted—it's not quite obvious at a first glance. Though not sure about placing, it's quite specific. Maybe even be placed in `Implementing Your Own DistributedActorSystem` section as it's mostly for systems to care about?

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+and only `distributed` methods may be called on it.
+
+> Note: Swift's distributed actor model features an unique bring-your-own-runtime approach,
+> meaning that there exist various specialized transport implementations (distributed actor systems)
+> which are specialized for specific transports (e.g. network, ipc, ...) or use-cases,
+> which all utilize the shared conceptual programming model.
+
+## Thinking in Distributed Actors
+
+In order to build distributed systems successfully you will need to get into the right mindset.
```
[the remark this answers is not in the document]
ktoso: I like this, thank you!

TSPL.docc/LanguageGuide/Distributed.md
```diff
@@ -0,0 +1,671 @@
...
+ 
+let player: Player = try Player.resolve(id: playerID, using: system)
+```
+
+This allows us to obtain a reference to a `Player` actor by asking the provided actor system to create a reference.
+
+The actor system may return a local or remote reference, however it should not perform asynchronous work such as 
+trying to confirm if the actor exists remotely or not. The resolve method should quickly return either a known
+local actor identified by the passed `id` or a remote reference if the actor may exist remotely.
+
```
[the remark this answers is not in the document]
ktoso: Hmmm, I agree we need to document this but maybe needs its own section with more info.

TSPL.docc/LanguageGuide/Distributed.md
ktoso: Codable is implicit here since the system we're on's ID is Codable -- I need to explain this in a section for its own I think.

TSPL.docc/LanguageGuide/Distributed.md
ktoso: That one's tricky, I would not want $ types in distributed method signatures but perhaps there's no workaround yet... I'll dig into this one

ktoso:
I'll revive this work, it's long overdue :)

ktoso:
We now have a space to document advanced things, so I did! https://github.com/swiftlang/swift/pull/88605

amartini51:
@ktoso Is there anything from this PR that should be incorporated into TSPL, or does the PR that adds articles supersede this one?

ktoso:
I totally missed the ping, yes! This is still valuable, let me brush it up finally. Thank you, Alex

Issue #242, opened 2024-01-11
Reconcile Keywords with Grammar


tadbyt:
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure#Keywords-and-Punctuation
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/summaryofthegrammar

### Description

The following keywords are not identified as such in the _Keywords and Punctuation_ subsection although they are given in various grammar productions in the _Summary of the Grammar_ section:

actor, any (with lowercase a), arch, arm, arm64, assignment, async, canImport, compiler, higherThan, i386, iOS, iOSApplicationExtension, Linux, lowerThan, macCatalyst, macOS, macOSApplicationExtension, nonisolated, os, package, safe, simulator, swift, targetEnvironment, tvOS, tvOSApplicationExtension, unsafe, visionOS, watchOS, watchOSApplicationExtension, Windows, x86_64

The specifics of which  keyword list they are missing from is given in the attachment.

throw and switch are reversed in _Keywords used in statements_
private and precedencegroup are reversed in _Keywords used in declarations_

### Correction

See attachment.
[Reconcile Keywords with Grammar.pdf](https://github.com/apple/swift-book/files/13896506/Reconcile.Keywords.with.Grammar.pdf)

PR #330, opened 2024-09-24
Alphabetize keywords in `LexicalStructure.md`

InsertCreativityHere:
While skimming the keyword reference, I noticed that 2 keywords were not listed in alphabetical order.
I know this is like, the most minor thing in the world, but it seems like everything else is properly ordered, so I figured... why not?

With this PR, all the keyword lists on this page would be in alphabetical order.

----

After doing a little more digging, there's actually an open issue for this! #242,
although it deals with more than just fixing the order of these two.


amartini51:
Thanks!  Yes, these lists are meant to be in alphabetical order.

amartini51 (approved):

amartini51:
Merged in commit eb096cd5.

PR #341, opened 2024-11-22
Fix type inference example in "Creating an Empty Array"

bohdany-cricut:
This pull request resolves an inconsistency in the "**Creating an Empty Array**" section of the Swift Language Guide under "**Collection Types**."

### Changes:
- Updated the example code by replacing `var someInts: [Int] = []` with `var someInts = [Int]()` to accurately demonstrate type inference, aligning with the accompanying explanatory text.

### Rationale:
The previous example did not demonstrate type inference, as the type `[Int]` was explicitly declared using `: [Int]`. However, the accompanying text states:  
> "Note that the type of the `someInts` variable is **inferred** to be `[Int]` from the type of the initializer."  

This was misleading because type inference did not occur in the provided example. The updated example clarifies the explanation and ensures consistency across the guide, aligning with similar sections such as "**Creating and Initializing an Empty Set**."

### DIFF
--- a/TSPL.docc/LanguageGuide/CollectionTypes.md
+++ b/TSPL.docc/LanguageGuide/CollectionTypes.md
@@ -74,7 +74,7 @@ You can create an empty array of a certain type
 using initializer syntax:
 
 ```swift
-var someInts: [Int] = []
+var someInts = [Int]()
 print("someInts is of type [Int] with \(someInts.count) items.")
 // Prints "someInts is of type [Int] with 0 items."
 ```
--- a/TSPL.docc/LanguageGuide/CollectionTypes.md
+++ b/TSPL.docc/LanguageGuide/CollectionTypes.md
@@ -70,9 +70,7 @@ and is used throughout this guide when referring to the type of an array.
 
 ### Creating an Empty Array
 
-You can create an empty array in Swift using two approaches,
-as shown in the following examples.
-
+You can create an empty array in Swift using two approaches.
 If the context already provides type information,
 such as a function argument or an already typed variable or constant,
 you can use an empty array literal,
--- a/TSPL.docc/LanguageGuide/CollectionTypes.md
+++ b/TSPL.docc/LanguageGuide/CollectionTypes.md
@@ -70,8 +70,34 @@ and is used throughout this guide when referring to the type of an array.
 
 ### Creating an Empty Array
 
-You can create an empty array of a certain type
-using initializer syntax:
+You can create an empty array in Swift using one of two approaches:
+
+If the context already provides type information,
+such as a function argument or an already typed variable or constant,
+you can create an empty array with an *empty array literal*,
+which is written as `[]`
+(an empty pair of square brackets):
+
+```swift
+var someInts: [Int] = []
+print("someInts is of type [Int] with \(someInts.count) items.")
+// Prints "someInts is of type [Int] with 0 items."
+```
+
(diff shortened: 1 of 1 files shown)


amartini51:
Thanks for taking the time to write this up in a PR!

This change looks like it’s only in an HTML `<!-- -->` comment, which isn't part of the rendered book.  The block that this PR changes is essentially a leftover from migration to DocC.  The build system TSPL used before DocC tested the code listings.  DocC doesn't support tested code listings, but we kept the tests during the export, so we would still have the hidden "expectation" lines if we ever try to resume code testing.

Looking at the [rendered](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/collectiontypes/#Creating-an-Empty-Array) version of the content, the example uses this form:

```swift
var someInts: [Int] = []
```

However, as you noted, the text and the code don’t match: The text talks about type inference, but the code has an explicit type in it.  This looks like this has been wrong since this content was added by commit a827c723391079ef48db67388061bea3ae7f79bd in 2015 — and I overlooked the issue in 2021 in commit 639bcc61dde12b16cb50fcd81885292bfd6a2883 when cleaning up initializer syntax throughout the book.

In this case, I think we might want to just remove the mention of type inference, since it would be pretty rare to create an empty array in code that has enough type information to infer the array’s type.  If we want to mention something about that, one example is passing an empty array to a function.

I'll be out of the office next week for Thanksgiving — if you'd like to try rewriting, take a look at your changes when I get back.  Otherwise, I'll push another commit to your branch after the holiday, changing the prose.

bohdany-cricut:
@amartini51 Thank you for taking the time to review my PR! I appreciate the context regarding the migration to DocC.

After carefully considering your feedback, I believe the best solution would be to update the example to use the actual initializer syntax as an alternative to the preferred array literal syntax. Here’s why I think this approach would address the concerns and improve the clarity of the guide:

**1. Clarifying "Initializer Syntax"**
The section currently states:

> You can create an empty array of a certain type using initializer syntax.

However, the example provided (`var someInts: [Int] = []`) does not use an initializer. Instead, it assigns an empty array literal to a variable with an explicitly declared type. This is misleading because it does not demonstrate what "initializer syntax" means.

Updating the section to explicitly show both methods - the preferred array literal syntax (`[]`) and the initializer syntax (`[Int]()`) - provides clarity and ensures completeness. Additionally, it’s worth noting that the "Collection Types" page doesn’t currently include any examples showcasing the array initializer syntax.

**2. Avoiding Redundancy**

The "Alternatively" example, which follows the initializer syntax section, states:

> Alternatively, if the context already provides type information, such as a function argument or an already typed variable or constant, you can create an empty array with an empty array literal, which is written as `[]` (an empty pair of square brackets):

This is illustrated with:

`someInts = []`

However, this example demonstrates the same concept as the first example: assigning an empty array literal ([]) to a variable. This creates redundancy without adding new information.

By explicitly stating that the array literal (`[])` is the preferred method and introducing the initializer syntax (`[Int]()`) as an alternative, the section would comprehensively cover both ways of creating an empty array without duplication.

amartini51:
Thanks for your revision and for your patience — I expected to come back to this PR sooner.  I've pushed a few minor changes to this branch.  This looks good to me now.

amartini51 (approved):

amartini51:
Reviewed & approved in person with Robert Olsen.

PR #342, opened 2024-11-28
Fix terminology for string literal in example explanation

bohdany-cricut:
This pull request addresses a terminology issue in the "[Overriding a Failable Initializer](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/#Overriding-a-Failable-Initializer)" section of the Swift Language Guide. The explanation incorrectly referred to `"[Untitled]"` from the example as a "string constant," which is not accurate. In Swift, `"[Untitled]"` is a string literal, not a constant.
```swift
super.init(name: "[Untitled]")!
```


amartini51 (approved):
This looks like a correct fix to me — thank you!

PR #345, opened 2024-12-20
Fix typo for nono-ptional with non-optional at Style.md

lamtrinhdev:
Fix typo for `nono-ptional` with `non-optional` at Style.md.

### DIFF
--- a/Style.md
+++ b/Style.md
@@ -164,7 +164,7 @@ See entry for *function*.
 
 ## non-optional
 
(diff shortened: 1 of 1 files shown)


amartini51:
Thanks for opening your a PR, to help make the Swift book better!  In this case, the spelling is intentional.  Reading it in context:

> Hyphenated to avoid the misreading as nono-ptional.

The spelling illustrates how "nonoptional" might be misread as starting with the syllables "no no".

PR #352, opened 2025-01-30
Add reference for noasync in @available [SE-0340]

Rich-Warren:
Fixes: rdar://142580634

### DIFF
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -274,8 +274,9 @@ including important milestones.
   ```
 
   You can use noasync availability with most declarations; 
-  however, you can't use it when declaring destructors.
-  The system must be able to call destructors from any context.
+  however, you can't use it when declaring deinitializers.
+  Swift must be able to call a class's deinitializers from any context, 
+  both synchronous and asynchronous.
 
 - The `message` argument provides a textual message that the compiler displays
   when emitting a warning or error about the use of a deprecated, obsoleted, 
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -237,7 +237,7 @@ including important milestones.
   ```
  
   This attribute tells the compiler to raise a build error 
-  when the symbol is used in an asynchronous context. 
+  when someone uses the symbol in an asynchronous context. 
   You can also use the `message` argument to provide additional information
   about the symbol.
 
@@ -249,7 +249,7 @@ including important milestones.
   ```
 
   If you can guarantee that your code uses a potentially unsafe symbol in a safe manner, 
-  You can wrap it in a synchronous function and call that function 
+  you can wrap it in a synchronous function and call that function 
   from an asynchronous context.
 
   ```swift
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -216,92 +216,53 @@ including important milestones.
 
   Because Swift concurrency can resume on a different thread
   after a potential suspension point,
-  you can't use elements like thread-local storage, locks, mutexes, or semaphores
-  across suspension points.
-  For example, accessing thread-local storage
-  on both sides of a suspension point could produce incorrect results.
+  using elements like thread-local storage, locks, mutexes, or semaphores
+  across suspension points can lead to incorrect results.
+
+  To avoid any errors, add an `@available(*, noasync)` attribute to the symbol's declaration:
 
   ```swift
-  let customThreadLog = "CustomThreadLog"
-  
-  // Reads the log for the current thread.
-  func readLog() -> String? {
-      return Thread.current.threadDictionary[customThreadLog] as? String
-  }
-  
-  // Add text to the log for the current thread.
-  func appendToLog(_ text: String) {
-      var log = readLog() ?? ""
-      log.append(text)
-      Thread.current.threadDictionary[customThreadLog] = log
-  }
-  
-  // This function modifies thread-local storage across a 
-  // potential suspension point.
-  func verify(name: String) async {
-      // This call appends the text to the current thread's log.
-      appendToLog("Verifying \(name).\n")
-      
-      // This is a potential suspension point.
-      let result = await coin.flip()
-
-      // This call might append the text to a different thread's log.
-      appendToLog("\(name) \(result ? "is" : "isn't") verified!\n\n")
-  }
-  ```
-  
-  In the previous code listing, 
-  the compiler doesn't have enough information to detect the problem. 
-  To prevent issues like this, 
-  add an `@available(*, noasync)` attribute to the symbol's declaration:
-
-  ```swift
-  // Reads the name for the current thread.
-  @available(*, noasync)
-  func readLog() -> String? {
-      return Thread.current.threadDictionary[customThreadLog] as? String
+  extension pthread_mutex_t {
+    
+    @available(*, noasync)
+    mutating func lock() {
+        pthread_mutex_lock(&self)
+    }
+    
+    @available(*, noasync)
+    mutating func unlock() {
+        pthread_mutex_unlock(&self)
+    }
   }
   ```
  
   This attribute tells the compiler to raise a build error 
   when the symbol is used in an asynchronous context. 
-  You can also use the `message` attribute to provide additional information
+  You can also use the `message` argument to provide additional information
   about the symbol.
 
   ```swift
-  // Reads the name for the current thread.
-  @available(*, noasync, message: "Use safeVerify(name:) instead.")
-  func readLog() -> String? {
-      return Thread.current.threadDictionary[customThreadLog] as? String
+  @available(*, noasync, message: "Migrate locks to Swift concurrency.")
+  mutating func lock() {
+    pthread_mutex_lock(&self)
   }
   ```
 
-  While declaring noasync availability prevents accidental, unsafe use of a symbol,
-  it can also prevent safe uses.
-  In some cases, the symbols can be used safely in specific situations.
-  If you can ensure that a particular use is safe in an asynchronous context,
-  wrap the symbol in a safe, synchronous function. 
-  You can then call the synchronous wrapper from your asynchronous code.
-  The compiler won't raise an error, because the `noasync` symbol 
-  isn't used directly in an asynchronous context.
+  If you can guarantee that your code uses a potentially unsafe symbol in a safe way, 
+  You can wrap it in a function and call that function 
+  from an asynchronous context.
 
   ```swift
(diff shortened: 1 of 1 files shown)


TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,105 @@ including important milestones.
...
   ```
   The *version number* consists of one to three positive integers, separated by periods.
+
+- The `noasync` argument indicates that the declared symbol can't be used directly
+  in an asynchronous context.
+
+  Because Swift concurrency can resume on a different thread
+  after a potential suspension point,
+  you can't use elements like thread-local storage, locks, mutexes, and semaphores
+  across suspension points. For example, accessing thread-local storage
```
amartini51, line 220:
```suggestion
  you can't use elements like thread-local storage, locks, mutexes, or semaphores
  across suspension points.
  For example, accessing thread-local storage
```

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,105 @@ including important milestones.
...
+  on both sides of a suspension point could produce unintended behaviors.
```
amartini51, line 221: Is there something more direct we can say than just "unintended behaviors"?  Maybe something like data races or bugs or incorrect results?

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,105 @@ including important milestones.
...
+  // between the appendToLog() calls.
+  func safeVerify(name: String) -> String {
+      appendToLog("Verifying \(name).\n")
+
+      // This is no longer a potential suspension point.
+      let result = coin.synchronousFlip()
+      
+      appendToLog("\(name) \(result ? "is" : "isn't") verified!\n\n")
+
+      // Multiple calls to safeVerify() may not occur on the same thread.
```
amartini51, line 299:
```suggestion
      // Multiple calls to safeVerify() can't occur on the same thread.
```

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,105 @@ including important milestones.
...
+  // Reads the name for the current thread.
+  @available(*, noasync, message: "Use safeVerify(name:) instead.")
+  func readLog() -> String? {
+      return Thread.current.threadDictionary[customThreadLog] as? String
+  }
+  ```
+
+  While the `noasync` argument prevents accidental, unsafe use of a symbol,
+  it can also prevent safe uses.
+  In many cases, `nonasync` symbols can still be used safely in specific situations.
```
amartini51, line 280: Do we need an English name for this kind of symbol, instead of referring to it by the syntax?

TSPL.docc/ReferenceManual/Attributes.md
Rich-Warren: I made some changes here to better describe the situation. Let me know if it still doesn't make sense.

TSPL.docc/ReferenceManual/Attributes.md
Rich-Warren: I changed the text to an English description to minimize duplicate references in the `noasync` reference.

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,108 @@ including important milestones.
...
+  ```swift
+  // Reads the name for the current thread.
+  @available(*, noasync, message: "Use safeVerify(name:) instead.")
+  func readLog() -> String? {
+      return Thread.current.threadDictionary[customThreadLog] as? String
+  }
+  ```
+
+  While declaring noasync availability prevents accidental, unsafe use of a symbol,
+  it can also prevent safe uses.
```
etcwilde, line 280: I think this bit could use some reworking.

`noasync` gives the compiler a mechanism to prevent you from accidentally hurting yourself. "can" is a dangerous word for a language. Pointers _can_ be used safely, but many times aren't. We could technically attempt to do something with structural analysis and tracking "opening" and "closing" calls to verify that no path has a potential suspension point, but that would likely be quite brittle. Instead, it's more principled to let the type-system handle this for us by requiring that the dangerous thing be used in a context where it can't be dangerous rather than trusting people to remember not to add suspension points between the opening and closing functions.

Sort of like how we only make the pointer to an object available inside of a closure (and you're not supposed to capture that pointer):
```swift
withUnsfePointer(to: thingy) { thingyptr in
  // ...
}
```

This was really designed with locks in mind where you have paired functions that take the lock and then unlock it. Between those two points is where the suspension point is dangerous. By taking a synchronous closure, we can ensure that it is impossible to accidentally introduce a suspension point.

```swift
func withLock(_ op: () -> ()) {
    myLock.lock()
    op()
    myLock.unlock()
}

withLock {
  // do locked thing
}
```

This is safe. There is a well-defined lifetime that the lock is held, and it guarantees that it's unlocked when you return from the synchronous operation. This helps prevent undefined behavior across the suspension point, as well as helping prevent deadlocks: (if you were to call this from the main thread, and then hop to the MainActor, for example).

I'm not sure that the logger example is the most compelling case for `noasync`. The operation that you're logging might not be something that can be synchronous (which is then why things like swift-log track things like task-ID's so that you can trace your progress while jumping across threads (or even to entirely different machines)).

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,69 @@ including important milestones.
...
+  ```
+
+  If you can guarantee that your code uses a potentially unsafe symbol in a safe way, 
+  You can wrap it in a function and call that function 
+  from an asynchronous context.
+
+  ```swift
+  // Using a wrapper ensures the mutex is locked
+  // and unlocked on the same thread.
+  func modifyProtectedData() -> Int {
```
etcwilde, line 258: I think we can continue expanding on the `lock` and `unlock` extension above and show them how to provide the safe wrapper that is made up of the unavailable functions by putting them in a synchronous function:

e.g adding a synchronous `withLock` that takes a synchronous closure:

```swift
extension pthread_mutex_t {
  @available(*, noasync, message: "Use withLock instead.")
  mutating func lock() {
    pthread_mutex_lock(&self)
  }

  @available(*, noasync, message: "Use withLock instead.")
  mutating func unlock() {
    pthread_mutex_unlock(&self)
  }

  mutating func withLock(_ op: () -> ()) {
    self.lock()
    op()
    self.unlock()
  }
}
```

Then if we still want to, we can show how to use it from the async function safely.

```swift
func downloadAndStore(key: String, dataStore: DataStore, dataLock: inout pthread_mutex_t) async -> Int {
  let content = await downloadContent(key)
  dataLock.withLock {
    dataStore[key] = content
  }
}
```

Thoughts?

Rich-Warren:
Let me know if this needs any additional changes. Thanks!

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,76 @@ including important milestones.
...
+    // Safely call the wrapper in an asynchronous context.
+    dataLock.withLock {
+      dataStore[key] = downloadContent()
+    }
+  }
+  ```
+
+  You can use noasync availability with most declarations; 
+  however, you can't use it when declaring destructors.
+  The system must be able to call destructors from any context.
```
etcwilde, line 278: I think that this bit could be a bit more consistent with the [deinitialization](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/deinitialization) section. Instead of destructors, deinitializers is probably better.

Something like, "Swift but be able to call the deinitializer for a class from any context, regardless of whether it is synchronous or asynchronous"?

etcwilde (approved):

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,77 @@ including important milestones.
...
+    }
+    
+    @available(*, noasync)
+    mutating func unlock() {
+        pthread_mutex_unlock(&self)
+    }
+  }
+  ```
+ 
+  This attribute tells the compiler to raise a build error
```
amartini51, line 239:
```suggestion
  This attribute tells the compiler to report an error 
```

Elsewhere in TSPL, we call these "compile-time errors"; the usual verb is "raise", although some places we have "results in a compile-time error".

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,77 @@ including important milestones.
...
+
+- The `noasync` argument indicates that the declared symbol can't be used directly
+  in an asynchronous context.
+
+  Because Swift concurrency can resume on a different thread
+  after a potential suspension point,
+  using elements like thread-local storage, locks, mutexes, or semaphores
+  across suspension points can lead to incorrect results.
+
+  To avoid any errors, add an `@available(*, noasync)` attribute to the symbol's declaration:
```
amartini51, line 222: Can we change the word "error" here?  It already has two meanings — a compile-time error, and a Swift `throw` style error — without us also using it to mean bug/mistake/etc.  Maybe "to avoid this problem", or "To mark code that uses these elements, add..." or similar?

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,77 @@ including important milestones.
...
+
+  If you can guarantee that your code uses a potentially unsafe symbol in a safe manner, 
+  You can wrap it in a synchronous function and call that function 
+  from an asynchronous context.
+
+  ```swift
+
+  // Provide a synchronous wrapper around methods with a noasync declaration.
+  extension pthread_mutex_t {
+    mutating func withLock(_ op: () -> ()) {
```
amartini51, line 259: Unless we're mirroring real API, let's use a full word.

```suggestion
    mutating func withLock(_ operation: () -> ()) {
```

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -210,8 +210,77 @@ including important milestones.
...
+
+  ```swift
+  @available(*, noasync, message: "Migrate locks to Swift concurrency.")
+  mutating func lock() {
+    pthread_mutex_lock(&self)
+  }
+  ```
+
+  If you can guarantee that your code uses a potentially unsafe symbol in a safe manner, 
+  You can wrap it in a synchronous function and call that function
```
amartini51, line 252: Typo "You" mid-sentence.  Probably don't need "you can" anyhow though.

```suggestion
  wrap it in a synchronous function and call that function 
```

Maybe "that wrapper function" instead of "that function", for clarity.

amartini51 (commented):
Minor wording nits only — overall this looks good.

I also see trailing whitespace in several places, which it looks like the web interface doesn't highlight.  You can run `git diff main... --check` to get a list of them.

PR #353, opened 2025-02-17
Add Android to #if os(…) list in Statements.md

marcprux:
Android is a valid OS for the`#if os(…)` check.

### DIFF
--- a/TSPL.docc/ReferenceManual/Statements.md
+++ b/TSPL.docc/ReferenceManual/Statements.md
@@ -1004,7 +1004,7 @@ conditions listed in the table below.
 
 | Platform condition | Valid arguments |
 | ------------------ | --------------- |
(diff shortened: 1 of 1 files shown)


amartini51 (changes requested):
I don't see Android in the list of supported platforms on Swift.org here <https://www.swift.org/platform-support/>.

As noted in the [in-source comments nearby](https://github.com/swiftlang/swift-book/pull/353/files#diff-f164dc1266910e33e476b5a82b427e888c8a59f1ba8daec18832a52078e5c526R1014) there are many platforms that don't appear in TSPL because they're supported by the compiler, but only unofficially or experimentally.

Do you think that list is outdated or incorrect?  Otherwise, we should close this PR without merging.  Thanks!

PR #368, opened 2025-05-19
Correct subscript setter name

an0:
Fix a typo


amartini51:
Tested as follows:

```
struct Bad {
    subscript (index: Int) -> Int {
       get { return 12 }
       set { print(value) }  // Error: Can't find 'value' in scope.
    }
}
struct Good {
    subscript (index: Int) -> Int {
       get { return 12 }
       set { print(newValue) }
    }
}
```

The current version doesn't compile.  With this PR's correction, the code compiles.

Issue #374, opened 2025-07-03
Docs state that accessor macros require the `names:` argument, but in reality they do not


daltonclaybrook:
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#attached

### Description

The documentation for the `@attached` macro attribute states the following:

>The peer, member, and accessor macro roles require a `names:` argument, listing the names of the symbols that the macro generates.

But in reality, accessor macros do not require the `names:` argument as they simply generate accessors for the stored property they're attached to. ([See discussion on the Swift forums](https://forums.swift.org/t/should-accessor-macros-declarations-require-a-names-argument-to-synthesize-getters-and-setters/70847))

### Correction

The sentence should be changed to:

>The peer and member macro roles require a `names:` argument, listing the names of the symbols that the macro generates.

vanvoorden:
Just to be safe… it might also be a good idea to open an issue in https://github.com/swiftlang/swift-syntax and confirm there that this is a legit behavior.

hborla:
It's a deliberate behavior that's documented in the attached macro proposal (this is also a behavior of the compiler and not of swift-syntax).

PR #375, opened 2025-07-03
Fix list of macro roles that need "names:" argument

daltonclaybrook:
This PR fixes a documentation bug stating that the `names:` argument is required for `@attached` macros with the `accessor` role. In reality, this argument is not required for accessor macros.

Fixes: https://github.com/apple/swift-book/issues/374

### DIFF
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -99,6 +99,10 @@ indicates the macro's role:
 
 The peer and member macro roles require a `names:` argument,
 listing the names of the symbols that the macro generates.
+The accessor macro role requires a `names:` argument if the
+macro generates a `willSet` or `didSet` property observer. An
+accessor macro that generates property observers can't add
+other accessors, because observers only apply to stored properties.
 The extension macro role also requires a `names:` argument
 if the macro adds declarations inside the extension.
 When a macro declaration includes the `names:` argument,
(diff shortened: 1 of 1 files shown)


amartini51:
@daltonclaybrook  Were you able to confirm that this is correct and expected behavior from Swift Syntax, as suggested on the GitHub issue?  If not, do you know who to talk to, or can I help you find a technical reviewer?  Thanks!

hborla:
@daltonclaybrook is correct; attached accessor macros do not require a `names` argument. The `names` argument is only required if the accessor macro generates property observers. This behavior is covered by SE-0389 in the section on accessor macros here: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0389-attached-macros.md#accessor-macros

TSPL.docc/ReferenceManual/Attributes.md
```diff
@@ -97,7 +97,7 @@ indicates the macro's role:
   on an extension, a type alias, or a type that's nested inside a function,
   or use an extension macro to add an extension that has a peer macro.
 
-The peer, member, and accessor macro roles require a `names:` argument,
+The peer and member macro roles require a `names:` argument,
```
hborla, line 100: I think we should add after this sentence something to the effect of

> The accessor macro role requires a `names:` argument if the
> macro generates a `willSet` or `didSet` property observer. An
> accessor macro that generates property observers can't add
> other accessors, because observers only apply to stored properties.

hborla: Thank you!

PR #381, opened 2025-08-19
Add a section on isolated conformances to the guide [SE-0470]

hborla:
This change introduces a section on global-actor isolated protocol conformances, which were introduced in [SE-0470](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0470-isolated-conformances.md).

Fixes: rdar://148361008

### DIFF
--- a/TSPL.docc/LanguageGuide/Concurrency.md
+++ b/TSPL.docc/LanguageGuide/Concurrency.md
@@ -1686,14 +1686,18 @@ protocol requirements can be used through type parameters or `any` types.
 
 #### Using Isolated Conformances
 
-A conformance requirement to `Sendable` allows generic code to send parameter
-values to concurrently-executing code.  If generic code accepts non-`Sendable`
-types, then the generic code can only use the input values from the current
-isolation domain. These generic APIs can safely accept isolated conformances
-and call protocol requirements as long as the caller is on the same global
-actor that the conformance is isolated to. The following code has a protocol
-`Dancer`, a class `Ballerina` with a main-actor isolated
-conformance to `Dancer`, and calls to the `Dancer.perform` requirement
+A conformance requirement to `Sendable`
+allows generic code to send parameter values to concurrently-executing code.
+If generic code accepts non-`Sendable` types,
+then the generic code can only use the input values
+from the current isolation domain.
+These generic APIs can safely accept isolated conformances
+and call protocol requirements
+as long as the caller is on the same global actor
+that the conformance is isolated to.
+The following code has a protocol `Dancer`,
+a class `Ballerina` with a main-actor isolated conformance to `Dancer`,
+and calls to the `Dancer.perform` requirement
 from a main-actor task and a concurrent task:
 
 ```swift
@@ -1819,7 +1823,8 @@ func performConcurrently<D: Dancer>(_ dancer: D) where D: Sendable {
 }
 ```
 
-The above code would admit data races if the conformance to `Dancer` was isolated,
+The above code would admit data races
+if the conformance to `Dancer` was isolated,
 because the implementation of `perform`
 may access global actor isolated state.
 To prevent data races,


DougGregor (commented):
Thank you for writing this!

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,307 @@ You can also use an unavailable conformance
...
+
+In the above code, the call to `performIfP` from a main-actor isolated task
+matches the isolation of the conformance, so the dynamic cast succeeds. The
+call to `performIfP` from a concurrent task happens outside the main actor,
+so the dynamic cast fails and `perform` is not called.
+
+#### Restricting isolated conformances in concurrent code
+
+Protocol requirements can be called through instances of conforming types
+and through metatype values. In generic code, a conformance requirement to
```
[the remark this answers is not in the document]
hborla: Ah, it might be the first place in the language guide. It's covered in the language reference here: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/types#Metatype-Type

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,307 @@ You can also use an unavailable conformance
...
+  }
+}
+```
+
+This allows the implementation of the conformance to use global actor
+isolated state while ensuring that state is only accessed from within the
+actor. Isolated conformances are also inferred for global actor isolated
+types. The following code example declares a conformance to `Equatable` for
+a main-actor isolated class, and Swift infers main-actor isolation for the
+conformance:
```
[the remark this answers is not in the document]
hborla: Yeah this section at least needs to mention the upcoming feature. I'm also open to making the example us main actor mode. Main actor mode isn't covered at all yet but the plan is to add it as part of the section on inference in https://github.com/swiftlang/swift-book/pull/371

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,355 @@ You can also use an unavailable conformance
 to suppress implicit conformance to a protocol,
 as discussed in <doc:Protocols#Implicit-Conformance-to-a-Protocol>.
 
+## Isolated Protocol Conformances
```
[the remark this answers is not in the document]
hborla: I think we will eventually generalize this feature to actors too. I'm not sure we need a note formatted like a note - we can clarify the intro to state that a conformance can be isolated to a global actor. We also probably don't need to repeat "global actor" afterward - a lot of the isolation rules aren't really specific to global actors. I like the section title as is because shorter section titles are easier to skip for in the outline for the chapter.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,355 @@ You can also use an unavailable conformance
...
+```
+
+Isolated conformances are only safe to use
+when the code is running on the global actor
+that the conformance is isolated to,
+so the dynamic cast only succeeds
+if the dynamic cast occurs on the global actor.
+If you declare a main-actor isolated conformance to `P`
+and call `performIfP` with an instance of the conforming type,
+the dynamic cast will only succeed
```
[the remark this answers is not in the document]
hborla: Adjusted the second sentence

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,355 @@ You can also use an unavailable conformance
...
+Protocol requirements can be used
+through instances of conforming types and through
+instances of the conforming types themselves
+called *metatype values*.
+In generic code,
+a conformance requirement to `Sendable` or `SendableMetatype`
+tells Swift that an instance or metatype value is safe to use concurrently.
+To prevent isolated conformances from being used outside of their actor,
+a type with an isolated conformance
+can't be used for a type that must also satisfy a conformance requirement
```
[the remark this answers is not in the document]
hborla: I attempted to clarify this :)

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,355 @@ You can also use an unavailable conformance
...
+types, then the generic code can only use the input values from the current
+isolation domain. These generic APIs can safely accept isolated conformances
+and call protocol requirement as long as the caller is on the same global
+actor that the conformance is isolated to. The following code has a protocol
+`P`, a class `C` with a main-actor isolated conformance to `P`, and
+call-sites to the `P.perform` requirement from a main-actor
+task and a concurrent task:
+
+```swift
+protocol P {
```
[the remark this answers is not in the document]
hborla: Yes definitely!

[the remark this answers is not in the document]
hborla: Updated to `Dancer` and `Ballerina` for now but we can change it to whatever.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+```
+
+### Data-Race Safety for Isolated Conformances
+
+Swift prevents data races for isolated conformances
+by ensuring that protocol requirements are only used
+on the global actor that the conformance is isolated to.
+In generic code,
+where the concrete conforming type is abstracted away,
+protocol requirements can be used through type parameters or `any` types.
```
amartini51, line 1685: I found this sentence hard to understand until I started reading the code listing.

By type parameters, I think this is referring to generic type parameters, right?  TSPL uses both terms, but mostly uses "type parameters" in the Generics chapter — in this context, it might be worth spelling out.

TSPL uses the term "boxed protocol types" rather than "`any` types" or "existential types".  Would you consider code that uses `any` to be a form of generic code, or should we adjust the wording at the start of the sentence?

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+from a concurrent task results in an error,
+because it would allow calling the main actor isolated implementation
+of `Dancer.perform` from outside the main actor.
+
+Abstract code that uses type parameters and `any` types
+can check whether a value conforms to a protocol
+through dynamic casting.
+The following code has a protocol `Dancer`,
+and a method `performIfDancer` that accepts a parameter of type `Any`
+which is dynamically cast to `any Dancer` in the function body:
```
amartini51, line 1744: Might be clearer to introduce the code listing with what details I should watch out for while reading it.  For example here, I think the salient point is the `value as? any Dancer` check, not (for example) the `Dancer` protocol.  We can add a paragraph after the code listing that walks through the code in smaller steps.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+from a main actor task
+is safe because it matches the isolation of the conformance.
+Calling `Dancer.perform` in generic code and on an `any Dancer` type
+from a concurrent task results in an error,
+because it would allow calling the main actor isolated implementation
+of `Dancer.perform` from outside the main actor.
+
+Abstract code that uses type parameters and `any` types
+can check whether a value conforms to a protocol
+through dynamic casting.
```
amartini51, line 1741: This new part of the discussion needs a clearer connection to the current topic.  How is this different from the existing discussion of `as?` in the Type Casting chapter?  I think you're re-introducing that syntax as buildup for the next example, showing when that `as?` cast succeeds?

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+
+    let dancer: any Dancer = ballerina
+    dancer.perform()
+}
+
+Task { @concurrent in
+    let ballerina = Ballerina()
+    startRecital(c)  // Error
+
+    let dancer: any Dancer = ballerina // Error
```
amartini51, line 1726: If there's a way to briefly state the error, let's include that in the comment.  Not all `// Error` comments in TSPL do this, but it's preferred when we can.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+
+Calling `Dancer.perform` in generic code and on an `any Dancer` type
+from a main actor task
+is safe because it matches the isolation of the conformance.
+Calling `Dancer.perform` in generic code and on an `any Dancer` type
+from a concurrent task results in an error,
+because it would allow calling the main actor isolated implementation
+of `Dancer.perform` from outside the main actor.
+
+Abstract code that uses type parameters and `any` types
```
amartini51, line 1739: We don't use the term "abstract code" elsewhere in TSPL — let's see if there's an existing term we can use here.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+        dancer.perform()
+    }
+}
+```
+
+Isolated conformances are only safe to use
+when the code is running on the global actor
+that the conformance is isolated to,
+so the dynamic cast only succeeds
+if the dynamic cast occurs on that global actor.
```
amartini51, line 1762: This seems to be the critical point — let's lead with it.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+
+#### Restricting Isolated Conformances in Concurrent Code
+
+Protocol requirements can be used
+through instances of conforming types and through
+instances of the conforming types themselves,
+called *metatype values*.
+In generic code,
+a conformance requirement to `Sendable` or `SendableMetatype`
+tells Swift that an instance or metatype value is safe to use concurrently.
```
amartini51, line 1804: This reads like a definition of `Sendable` but I think it's about the `where D: Sendable` clause in the code below?  We can check terminology here too — I see "conformance requirement" only twice in TSPL, and in both places it refers to the requirements that a type must implement in order to conform to a protocol.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+    Task { @concurrent in
+        dancer.perform()
+    }
+}
+```
+
+The above code would admit data races
+if the conformance to `Dancer` was isolated,
+because the implementation of `perform`
+may access global actor isolated state.
```
amartini51, line 1829: I'm having a hard time with this sentence, probably because of the irrealis mood on the verbs (would admit, was isolated, may access).  Are you saying that the code above is safe, and then describing why by telling me about something that can't happen?  Or are you describing a scenario where the code has possible data races?

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+indicates that the metatype of a type conforms to `Sendable`,
+which allows the implementation to share metatype values in concurrent code:
+
+```swift
+protocol Dancer {
+    init()
+    func perform()
+}
+
+func performConcurrently<D: Dancer>(n: Int, for: D.Type) async where T: SendableMetatype {
```
amartini51, line 1867: I think the `T` is just a leftover from a previous version that had different type names?

```suggestion
func performConcurrently<D: Dancer>(n: Int, for: D.Type) async where D: SendableMetatype {
```

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+                dancer.perform()
+            }
+        }
+    }
+}
+```
+
+Note that `Sendable` requires `SendableMetatype`,
+so an explicit conformance to `SendableMetatype` is only necessary
+if the type is non-`Sendable`.
```
amartini51, line 1905: This information seems like it should come earlier.

TSPL.docc/LanguageGuide/Concurrency.md
```diff
@@ -1591,6 +1591,370 @@ You can also use an unavailable conformance
...
+```swift
+@MainActor class Ballerina: @MainActor Dancer {
+    init() { /* use main actor state */ }
+    func perform() { /* use main actor state */ }
+}
+
+let items = performConcurrently(n: 10, for: Ballerina.self)  // Error
+```
+
+##### Protocols That Require `Sendable` or `SendableMetatype`
```
amartini51, line 1921: We can't use H5 or code voice in headings.

amartini51 (commented):
Thanks for writing all of this!  I've marked a bunch of specific places throughout the diff — and some general comments below.

Let's expand the transitions and explanations.  Some sections have a main point too far into the section, and we can elevate that to the introductory paragraph. Several of the code listings would benefit from a paragraph after them that takes the reader through what the code's doing in more detail.

Some subsections don't need as many headings, and headings don't need to be nested as deeply.  (Although there's no issue with short sections when they're called for.)  Let's try to fit more with the level and length of headings in the existing book.  Some of the choices around sectioning in the existing content are based on what DocC handles well, some are based on editorial feedback, and some are just convention that we can vary.  Here's an outline of this chapter:

* Defining and Calling Asynchronous Functions                                                                                                                        
* Asynchronous Sequences                                                                                                                                             
* Calling Asynchronous Functions in Parallel                                                                                                                         
* Tasks and Task Groups                                                                                                                                              
    - Task Cancellation                                                                                                                                              
    - Unstructured Concurrency                                                                                                                                       
* Isolation                                                                                                                                                          
* The Main Actor                                                                                                                                                     
* Actors                                                                                                                                                             
* Global Actors                                                                                                                                                      
* Sendable Types                                                                                                                                                     
* Isolated Protocol Conformances  **<——— NEW**
    - Declaring an Isolated Conformance                                                                                                                              
    - Inferring an Isolated Conformance                                                                                                                              
    - Data-Race Safety for Isolated Conformances                                                                                                                     
        + Using Isolated Conformances                                                                                                                                
        + Restricting Isolated Conformances in Concurrent Code                                                                                                       
            + Protocols That Require `Sendable` or `SendableMetatype`                                                                                                

Prefer active voice and present tense over passive voice and future tense.  We'll catch this during the editorial pass too.

PR #408, opened 2025-10-27
Make use of existing syntactic category name

mattmassicotte:
I've been reviewing the formal grammar, and I noticed some duplication. I know this is a judgment call, and I also know that this doesn't keep the rules quite as isolated. However, the existing rule definition also doesn't fit well with the style guide anyways, so it seems like it was at least worth a discussion.

For context, the existing definition was:

```
> *decimal-digit* → Digit 0 through 9 \
```


mattmassicotte:
I've read through the guide a few times, and I do not see how `Digit 0 through 9` is acceptable syntax compared to using an explicit alternation with literals. But perhaps this is an intentional relaxation to make reading easier?

PR #410, opened 2025-11-06
Update the URL for Swift Atomics in a comment

kkebo:
https://github.com/apple/swift-se-0282-experimental no longer exists. Instead, this change replaces it with https://github.com/apple/swift-atomics.


heckj:
The link isn't exposed in content, but it is nice to have the correct reference if the comment remains

PR #417, opened 2025-12-02
Add print statement in control flow example

tomek7667:
The [guided tour](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/guidedtour/) example doesn't specify print after the greeting, so the experiment that follows won't be valid:

> Change optionalName to nil. What greeting do you get? Add an else clause that sets a different greeting if optionalName is nil.

### DIFF
--- a/TSPL.docc/GuidedTour/GuidedTour.md
+++ b/TSPL.docc/GuidedTour/GuidedTour.md
@@ -423,6 +423,7 @@ var greeting = "Hello!"
 if let name = optionalName {
     greeting = "Hello, \(name)"
 }
+print(greeting)
 ```
 
 <!--


amartini51:
Thanks for pointing this out!  The examples throughout the tour use "bare" expressions in several places, expecting that the reader is running the code in a REPL or playground, so the results of function calls are visible even without printing.  Older versions of the book included a link to download the chapter as a playground, which was lost when moving to DocC (see https://github.com/swiftlang/swift-book/issues/33).  However, looking again today, DocC now supports a [`@CallToAction`](https://www.swift.org/documentation/docc/calltoaction) directive, which should let us restore that download-playground functionality.

With that in mind, I think we're better off fixing the reader in general, instead of spot-fixing one specific example.  Or, if it turns out we can't use `@CallToAction`, the tour should be revised to include explicit `print()` everywhere the code produces a result that otherwise wouldn't be visible.

amartini51:
I've copied some of my notes above from December 8 to the issue tracking a long-term fix, and let's go ahead and merge the spot-fix now.

PR #443, opened 2026-02-16
Add documentation for body macros [SE-0415]

literally-anything:
Body macros are a stable language feature, but they are't mentioned anywhere except the evolution proposal.  This adds `body` as a possible type of attached macro in the `@attached` attribute section.

Fixes: rdar://171280404

### DIFF
--- a/TSPL.docc/ReferenceManual/Attributes.md
+++ b/TSPL.docc/ReferenceManual/Attributes.md
@@ -97,6 +97,14 @@ indicates the macro's role:
   on an extension, a type alias, or a type that's nested inside a function,
   or use an extension macro to add an extension that has a peer macro.
 
+- term Body macros:
+  Write `body` as the first argument to this attribute.
+  The type that implements the macro conforms to the `BodyMacro` protocol.
+  These macros modify or replace the body of the function that they are attached to.
+  If the function has no body,
+  the macro can generate one from scratch.
+  Only one body macro is allowed on a function at a time.
+
 The peer and member macro roles require a `names:` argument,
 listing the names of the symbols that the macro generates.
 The accessor macro role requires a `names:` argument if the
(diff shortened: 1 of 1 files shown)


amartini51 (commented):
Thanks for taking the time to add this to the documentation!  Sorry for a bit of delay last week, while I was working on some other documentation.

Reading this PR and SE-0415, I think there are two things we can check on.  Tagging @DougGregor as the proposal author, for technical review — maybe Doug can answer these.
 
Body macros can go on property accessors, like `get` and `set` — not just functions and methods.  Is that specific to body macros, or is it also true of all attached macros?  Let's add it to the reference either here, or for attached macros in general.

The proposals notes that body of a function marked with this macro must be syntactically valid Swift code, although it doesn’t need to typecheck.  The "Macros" chapter in TSPL says that normally code marked with macros does need valid type information:

> Both the input to a macro and the output of macro expansion are checked to ensure they’re syntactically valid Swift code. Likewise, the values you pass to a macro and the values in code generated by a macro are checked to ensure they have the correct types.

If the "no type checking" rule is specific to body macros, let's add it here — otherwise, at the relevant spot in the reference.

literally-anything:
Sorry, it took me a bit to get back to this.

I looked into the issues that you mentioned, and from what I can tell, body and preamble macros are allowed on constructors, destructors, functions, and accessors. So far they are the only macros that are allowed on accessors, so I added that.

On the type checking front, from what I can tell, all macros are evaluated during type checking, so it is never really type checked beforehand. But body macros are the first macro that can directly replace existing code, so I think this is just the first case where the order is noticeable. I’ve added a sentence about this to the page as well.

amartini51:
Re-reviewed now that I'm back from new-parent leave.  This looks good to me.  I'll follow up with @DougGregor to review for technical accuracy.

DougGregor (approved):
Looks good, thank you!

amartini51 (approved):

PR #454, opened 2026-03-26
extracts the grammar generation step into it's own script

heckj:
This lets us run this setup script with an external repository's (swiftlang/docs) build script to do the expected prep work for the DocC catalog in order to combine it with other DocC catalogs for a combined archive of multiple doc sets together.

work needed to support https://github.com/swiftlang/docs/pull/72, ultimately to support https://github.com/swiftlang/docs/issues/24

### DIFF
--- a/bin/generate-grammar
+++ b/bin/generate-grammar
@@ -1,0 +1,37 @@
+#! /bin/bash
+
+# This source file is part of the Swift.org open source project
+#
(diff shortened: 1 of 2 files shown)


invalidname (approved):
I haven't run it and not sure how I would, but it seems like a clean enough refactor.

PR #465, opened 2026-06-08
Fix missing period in GuidedTour experiment block

iTanmayTrivedi:

Fixes: rdar://####


amartini51:
Thanks for your contribution — however, adding a period in the middle of a sentence is incorrect.  The text is correct as-is:

> Add an `else` clause that sets a different greeting if `optionalName` is `nil`.

PR #470, opened 2026-06-23
Fix typo: writeable -> writable

wjdtjq6:
The protocol property requirements section uses 'writeable', while the rest of the book uses 'writable' (for example, WritableKeyPath in Attributes). This corrects the spelling for consistency.


amartini51 (approved):
Thanks!  Looks like this spelling has been here since commit e630853c in 2014 — I also confirmed that there aren't any other places to fix.  The remaining matches for `*eable` in the book are words like "cacheable" and "replaceable".

wjdtjq6:
Thank you for reviewing and merging this! I appreciate the extra check for other matches as well.

PR #473, opened 2026-07-12
Fix function type grammar for labeled parameters

### LINKED ISSUE #469: funtion-type-argument does not allow for external parameter names
### Location

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/summaryofthegrammar/

### Description

The following does not parse with TSPL grammar:

`typealias MyAlias = (_ a: Int, _ b: Double, _ c: Bool, _ d: String) -> Bool`

Because function-type-argument is missing the equivalent of external-parameter-name:

`function-type-argument -> attributes? parameter-modifier? type | argument-label type-annotation `

### Correction

The fix could be:

```
function-type-argument -> attributes? parameter-modifier? type | external-argument-label? local-argument-label type-annotation
external-argument-label -> identifier
local-argument-label -> identifier
```

somiljain2006:
Fixes: https://github.com/apple/swift-book/issues/469

Update the function type grammar to support function parameters with external and local argument labels. This fixes the grammar so it correctly accepts function type syntax such as `(_ a: Int, _ b: Double) -> Bool`, bringing the TSPL grammar in line with the syntax accepted by the Swift compiler.

### DIFF
--- a/TSPL.docc/ReferenceManual/Types.md
+++ b/TSPL.docc/ReferenceManual/Types.md
@@ -514,8 +514,9 @@ see <doc:MemorySafety>.
 > *function-type-argument-clause* → **`(`** *function-type-argument-list* **`...`**_?_ **`)`**
 >
 > *function-type-argument-list* → *function-type-argument* | *function-type-argument* **`,`** *function-type-argument-list* \
-> *function-type-argument* → *attributes*_?_ *parameter-modifier*_?_ *type* | *argument-label* *type-annotation* \
-> *argument-label* → *identifier*
+> *function-type-argument* → *attributes*_?_ *parameter-modifier*_?_ *type* | *external-argument-label*_?_ *local-argument-label* *type-annotation* \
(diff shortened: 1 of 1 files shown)


somiljain2006:
@amartini51 can you review this pr?

amartini51:
Thanks for taking a look at this part of the grammar!

In [the reference](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/declarations#Parameter-Names), I see the terms the terms "parameter name" and "argument label" — I don't see the term "external argument label" or "local argument label" used elsewhere.  Is this difference intentional?  If so, why?

What's the reason for defining "local argument label" as a new category, rather than re-using "identifier"?

somiljain2006:
Thanks for the review. My intention in introducing a separate nonterminal wasn't to change the grammar accepted by the language. Rather, it was to distinguish the two identifiers in a labeled function type for readers, similar to how the declaration grammar distinguishes external-parameter-name and local-parameter-name, even though those ultimately expand to identifier.

PR #474, opened 2026-07-22
Suppress the language version from the module name

heckj:
This hides the (6.4 beta) extension on the module name from the name that's presented in the combined view of Swift documentation (a merged set of DocC catalogs).

The "easier" method may be to update the title itself, but that also removes it from interior content as well.

### DIFF
--- a/TSPL.docc/The-Swift-Programming-Language.md
+++ b/TSPL.docc/The-Swift-Programming-Language.md
@@ -4,6 +4,7 @@ Write safe, fast, expressive code with a modern, general-purpose language.
 
 @Metadata {
   @TechnologyRoot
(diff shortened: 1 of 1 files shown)


amartini51:
Is there any change when building TSPL on its own?  If not, I think we're good to make this change in preparation for the new publication method.

heckj:
local builds remain identical, it highlights the version in the catalog title

amartini51:
Since I already have a shell script to diff builds of TSPL, I also confirmed that this PR doesn't change the output of `bin/publish-book` compared to what's on 'main'.

amartini51 (approved):

PR #485, opened 2026-08-31
Add legacy section redirects

### LINKED ISSUE #72: Set up redirects on Swift.org
Because of the way Swift-DocC works, the URL we publish to must change from <https://docs.swift.org/swift-book> to <https://docs.swift.org/documentation/swift-book>.  We need to set up server-side redirects on Swift.org so that links to chapters continue to work.  These should be relatively straightforward.

In addition, TSPL specifies a stable ID for each heading's anchor.  This is used for on-page navigation within the book, and frequently used outside of the book to provide a link directly to a specific heading.  Because the DocC dialect of markdown doesn't allow us to specify an anchor's ID, all of these stable IDs will be destabilized and instead derived from the text of the section heading.  (See also https://github.com/apple/swift-docc/issues/345)

For the second issue, we don't believe the web server can inspect the URL's fragment and offer different redirect options.  It may be possible to use a bit of JavaScript on a "trampoline" page, located at the old URL, to redirect to the new URL.

minjun0317:
## Summary

Adds redirects for legacy section links that now point to the corresponding DocC pages.

- maps the removed dictionary key-type section to its current fallback
- adds redirects for Control Flow and its renamed for-in and repeat-while sections
- adds redirects for the Deinitialization and Enumerations chapters

Fixes #72

## Validation

- Ran `swiftc -typecheck bin/redirects/redirect.swift`.


amartini51 (changes requested):
Validating that the Swift file typechecks is useful, but doesn't confirm this PR's actual changes.  For example, how did you confirm that these redirects link the correct old and new URLs?  This PR adds several redirects (which is great progress!) but doesn't resolve all of them so we shouldn't mark this as "resolving" the issue.

```
grep 'docc: ""' bin/redirects/redirect.swift | wc -l
```

